PR #698, opened 2014-07-20
Fix bug in dynare_estimation_init.m if steady state does not solve

JohannesPfeifer:
There was no check in dynare_estimation_init.m whether the steady state solved, but the noconstant option was set nevertheless. This gave rise to cases where a steady state file returned 0 for the observables, but the steady state did not exits. Due to later using use_calibration, this problem with the initial values was not detected, but the observation equation featured no constant although it should. The solution is to move the use_calibration option before the first steady state computation and then issue an error if the parameterization does not work

### DIFF
--- a/matlab/dynare_estimation_1.m
+++ b/matlab/dynare_estimation_1.m
@@ -107,29 +107,6 @@ if options_.dsge_var
     end
 end
 
-%check for calibrated covariances before updating parameters
-if ~isempty(estim_params_)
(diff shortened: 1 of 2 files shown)


JohannesPfeifer:
@MichelJuillard @stepan-a Any thoughts on this? This is one of the oldest bugfixes around and #852 still needs to be made compatible with this fix.

PR #710, opened 2014-08-14
Fix bug in dynare_estimation_init.m with analytic_derivation where param...

JohannesPfeifer:
...eter vector was not set

Before testing whether the steady state file changes parameter values, the parameter values were not set, leading to subsequent crashes


JohannesPfeifer:
@rattoma If I am correctly informed, you are the person responsible for dealing with analytic derivation. Can you confirm this fix?

rattoma:
Yes, I think this is correct. thanks a lot

JohannesPfeifer:
@stepan-a Could you pull this?

PR #732, opened 2014-09-17
Prior checks

JohannesPfeifer:
Deal with cases where initial prior density if Inf


stepan-a:
Manually merged (also fixed one typo)

PR #737, opened 2014-09-20
Fix bug in pm3 for filtered variables that led to crashes

JohannesPfeifer:
Check for size of array failed with singleton dimension


stepan-a:
Manually merged.

PR #839, opened 2015-01-21
Bug correction in splitting the jacobian into blocks: the vector variabl...

FerhatMihoubi:
...e_2_block is valid only for endogenous variables.

### DIFF
--- a/preprocessor/DynamicModel.cc
+++ b/preprocessor/DynamicModel.cc
@@ -3253,12 +3253,13 @@ DynamicModel::collect_block_first_order_derivatives()
       int var = symbol_table.getTypeSpecificID(getSymbIDByDerivID(it2->first.second));
       int lag = getLagByDerivID(it2->first.second);
       int block_eq = equation_2_block[eq];
-      int block_var = variable_2_block[var];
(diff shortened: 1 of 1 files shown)


JohannesPfeifer:
@FerhatMihoubi Does this also affect Dynare 4.4? If yes, we should cherry pick the fix and add the problem to the bug list.

houtanb:
Thanks @JohannesPfeifer . I cherry-picked the change to 4.4. It's in 2a0aeb6a6f3b3adf2a0a5daf7dced335084c690e

JohannesPfeifer:
@houtanb Thanks. But what is the underlying bug for the bug report page? Does this solve the issue reported in my email to @FerhatMihoubi on December 8th?

houtanb:
@JohannesPfeifer I don't remember your email, but it's due to a problem where an array entry was accessed that was sometimes not set, causing the preprocessor to crash.

JohannesPfeifer:
@houtanb Then it's something different. I opened a ticket for the other issue. As you seem to understand the problem, could you add this to the bug list?

Issue #780, opened 2014-11-11
Deal with options_.nobs and options_.first_obs in master


JohannesPfeifer:
With the new data interface, these fields are not always set, leading to crashes in different places. For example, recursive forecasting still relies on those fields. See also afb5be206741457a9b6383fa004bb3f208ed5d94 
We need to decide on how to treat them. Are these field properties of the dataset to be moved into `dataset_`? Or are they to be used as options in other commands?

JohannesPfeifer:
@stepan-a I understand that you have a lot to deal with, but my tickets rely on this bug being fixed. If you find 5 minutes, could you please just fix this along the lines outlined above.

stepan-a:
@JohannesPfeifer, I still have a problem with your proposed change. I tried to run a recursive estimation with fs2000.mod with the old data interface and it ran perfectly fine... So I do not understand the problem you are trying to fix. Can you point me to a mod file where the issue you are referring pops up?

JohannesPfeifer:
@stepan-a The issue in recursive estimation is gone by now. What remains is the problem in e.g. ```dyn_forecast``` and ```filt_mc_```. Just take the fs2000.mod and add the ```forecast``` option: 
```
/*
 * This file replicates the estimation of the cash in advance model described
 * Frank Schorfheide (2000): "Loss function-based evaluation of DSGE models",
 * Journal of Applied Econometrics, 15(6), 645-670.
 *
 * The data are in file "fsdat_simul.m", and have been artificially generated.
 * They are therefore different from the original dataset used by Schorfheide.
 *
 * The equations are taken from J. Nason and T. Cogley (1994): "Testing the
 * implications of long-run neutrality for monetary business cycle models",
 * Journal of Applied Econometrics, 9, S37-S70.
 * Note that there is an initial minus sign missing in equation (A1), p. S63.
 *
 * This implementation was written by Michel Juillard. Please note that the
 * following copyright notice only applies to this Dynare implementation of the
 * model.
 */

/*
 * Copyright (C) 2004-2010 Dynare Team
 *
 * This file is part of Dynare.
 *
 * Dynare is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * Dynare is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
[...]

Issue #804, opened 2014-11-26
Factorize saving of smoother results used in dynare_estimation_1 and evaluate_smoother


JohannesPfeifer:
In ```dynare_estimation_1``` both the calibrated smoother and ML/MCMC without replications write the smoother results to oo_. The same holds true for ```evaluate_smoother.m```. I would propose to outsource this code to a separate function that is easier to maintain (in particular when working on #679). This would also make sure that the saving is consistent. Currently, I don't think the calibrated smoother saves all necessary fields. 

If agreed upon, I would volunteer to do this.

JohannesPfeifer:
Closed with #1145

Issue #803, opened 2014-11-25
Be more flexible with calib_smoother command


JohannesPfeifer:
Currently, we do not allow ```prefilter```, ```loglinear```, and ```first_obs```. I think they are all valid options that can be easily handled within the current framework. The first step would be to add the preprocessor interface allowing these options.

PR #852, opened 2015-03-01
Trend prefiltering fixes

### LINKED ISSUE #255: Check for consistent use of dataset_.descriptive.mean vs. bayestopt_.mean_varobs in smoother and estimation
Currently, both structures are still present. prior_posterior_statistics.m sets 

bayestopt_.mean_varobs = dataset.descriptive.mean';

to have both. Thus, use should be made consistent and storage be confined to one structure. As the mean of the data belongs to the dataset and not the estimation, the dataset_ structure is better suited than bayestopt_.

### LINKED ISSUE #679: Fix various bugs related to detrending and prefiltering
- ```dyn_forecast``` adds unlogged steady state if ```loglinear``` is used
- ```dynare_estimation_init``` sets the ```noconstant``` based on the unlogged steady state if ```loglinear``` is used
- when trends are specified as a function of deep parameters, the values are not correctly updated during estimation due to using the base-workspace parameter values and not the updated local ones
- when using trends with the ```prefilter``` option, the mean shift due to the trend is not accounted for
- when using ```first_obs>1```, the higher trend starting point is not taken into account (leads also to problems in recursive forecasting)

JohannesPfeifer:
Mostly fixes #679 and closes #255, #803, and #804.
Still to be done: 
- document new options once implementation is OKed
- check treatment of various options and trends in smoother used in ```imcforecast.m``` 
- Check treatment of these issues in ```identification```
- Adjust ```convert_dyn_45_to_44``` to reflect these changes


JohannesPfeifer:
Also fixes #780

stepan-a:
@JohannesPfeifer Thanks for splitting your changes into multiple commits. I still have a doubt about c32b1ec43a7047fe3bebf85913c56c649a972991 (#780). Whenever possible we should avoid to version binary files (because it makes no sense to track history of a binary file and because we need to know where the data are coming from to understand the test). In 2d3a80c7fdaa32c6cf57ed09ca285f4bb511e697 could you replace the ```*.mat``` files by matlab scripts (or mod files) generating these files?

JohannesPfeifer:
@stepan-a I was also reluctant to add the binary mode-files. But it turned out that doing mode-finding with ```mode_compute=9``` or ```mode_compute=10``` took a long time given the size of the data. That is why I opted to include the saved mode-files. Given that there are 16 mod-files that require mode-finding, this would add about 8 hours to the run-time of the test-suite. Thus, I thought it would be preferable to provide the mode (they were generated by the same mod-files).

stepan-a:
@JohannesPfeifer Why would we need to use ```mode_compute=9, 10``` for testing that there is no problem with profiteering of the data?

JohannesPfeifer:
I basically test whether the estimation runs away from the correct previous values. But you are correct that it makes more sense to start at the calibrated true value and run mode_compute=4. I changed it that way

JohannesPfeifer:
@MichelJuillard @stepan-a From my perspective, this is the last big bugfix remaining for 4.5, but I guess there are still many kinks to be smoothed out. I now rebased it to reflect all the pull requests committed today. Any feedback is highly appreciated

stepan-a:
Done in #1145

PR #856, opened 2015-03-12
Fixed issues related to newratflag.

rattoma:
This option relates to alternative numerical hessian computations:

optim=('Hessian',1) is the default dynare numeric Hessian
optim=('Hessian',0) is the outer product gradient
optim=('Hessian',2) is the 'mixed' outer product gradient, where diagonal elements are obtained using second order derivation formula and outer product is used for correlation structure.

Both 0 and 2 cases require univariate filters, to ensure using maximum number of individual densities.
Both 0 and 2 are quicker than default dynare numeric Hessian, but provide decent starting values for Metropolis for large models (option 2 being the best out of the two options)

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -4855,7 +4855,9 @@ Uses Chris Sims's @code{csminwel}
 
 @item 5
 Uses Marco Ratto's @code{newrat}. This value is not compatible with non
-linear filters or DSGE-VAR models
+linear filters or DSGE-VAR models.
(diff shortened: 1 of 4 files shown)


JohannesPfeifer:
@rattoma Could you please also add this to the manual? Currently, there is no info on the values that the ```optim```-option takes for ```mode_compute=5```

rattoma:
@JohannesPfeifer: done.

rattoma:
I added the documentation to the same pull request. thank you.

PR #865, opened 2015-04-01
Diffuse filter and smoother

rattoma:
- harmonize tolerance criteria for diffuse likelihood to diffuse smoother
- avoid overwriting global restrict var list in smoother
- fix numerical issues in `schur_statespace_transformation`

### DIFF
--- a/matlab/DsgeSmoother.m
+++ b/matlab/DsgeSmoother.m
@@ -69,9 +69,10 @@ M_ = set_all_parameters(xparam1,estim_params_,M_);
 %------------------------------------------------------------------------------
 % 2. call model setup & reduction program
 %------------------------------------------------------------------------------
-oo_.dr.restrict_var_list = bayestopt_.smoother_var_list;
-oo_.dr.restrict_columns = bayestopt_.smoother_restrict_columns;
-[T,R,SteadyState,info,M_,options_,oo_] = dynare_resolve(M_,options_,oo_);
+DynareResults = oo_;
+DynareResults.dr.restrict_var_list = bayestopt_.smoother_var_list;
+DynareResults.dr.restrict_columns = bayestopt_.smoother_restrict_columns;
+[T,R,SteadyState,info,M_,options_,DynareResults] = dynare_resolve(M_,options_,DynareResults);
 bayestopt_.mf = bayestopt_.smoother_mf;
 if options_.noconstant
     constant = zeros(vobs,1);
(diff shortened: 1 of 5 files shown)


stepan-a:
@rattoma I understand the that tolerance criteria for diffuse filter and diffuse smoother should be the same. But looking at the diffuse smoother routines the tolerance parameter ```crit1``` is set equal to 1e-8 (not 1e-6 as you wrote in commit 511ee5378b9cdde9bbad07b83b924b252afe06de). Also why should we use a hard coded value? We could use ```options_.kalman_tol``` (or add a new option if the default value of ```kaman_tol``` is too small as you suggest). Also, looking at the diffuse filter code (but this not in your commit), it is not clear to me why we should use ```crit1``` when testing the rank of ```Pinf``` and ```kalman_tol```when testing the condition number of ```Finf```.

rattoma:
@stepan-a. Concerning the hardcoded `crit1` I would very much like to add a new option specific for diffuse step, with default 1.e-8.
Concerning the tolerance for `Finf`: 

- in the multivariate diffuse filter [which by the way is almost never used in practice, since usually we have a mixture of unit roots and stationary variables in observation equations], you are right that we should always use `crit1`, also for testing `Finf`: I did a mistake there.
- in the univariate filter, in turn, it makes sense to use `kalman_tol` for `Finf` since an individual diagonal term can become small enough to need a tighter tolerance, especially for models with a lot of unit roots [i.e. the univariate diffuse algorithm while killing unit roots one by one can lead to small enough non-zero diagonal terms]. Viceversa, the much bigger `Pinf`matrix requires the courser `crit1`to avoid numerical noise affecting its rank calculations [I am talking about matrix sizes hundreds x hundreds: for small models this discussion is irrelevant]

By the way, I was the one introducing `crit1`in the smoother a long time ago, which made the diffuse steps work OK for so many years [in the smoother the noise issues are bigger than for the filter, due to the larger number of zero eigenvalues, I address this issue in commit 0336d5e].

stepan-a:
@rattoma Ok, I understand now. I will change the tolerance parameter in the multivariate diffuse filter, and add an option. I am short of ideas for the name, would ```kalman_smoother_tol``` be a good name for you?

I also noticed that the hard coded values for ```crit1``` are different in the multivariate and univariate diffuse filter... Is this difference intentional? Should we have two options (for instance ```kaman_smoother_tol.univariate``` and ```kaman_smoother_tol.multivariate```)?

rattoma:
for the option name I would use `diffuse_kalman_tol` since it applies to both filter and smoother.

for the difference in `crit1`between multi and univ folters I was just sloppy sorry.

thanks a lotg

stepan-a:
@rattoma Ok, I do that.

stepan-a:
Merged manually (fixed bug in 363c4a61222103fda22adff9bdb1115a6073428a and added new option).

PR #873, opened 2015-04-06
Optimizers: Extensions and bugfixes

### LINKED ISSUE #814: Decide upon what to do stale mode_compute options
In the process of factorizing the optimization calls (see #800), we should clean up. Currently, ```mode_compute==2,101,102``` are broken/not supported anymore. 2 and 102 are non-existing simulated annealings. I would propose to drop 102 and replace 2 by Matlab's 
```simulannealbnd``` from the Global Optimization Toolbox. That way we would have at least one simulated annealing available.

I have no clue why we have the undocumented 101 which is ```solveopt``` and would suggest to drop it.

JohannesPfeifer:
- Adds solvopt and simulated annealing from public domain: http://www.dynare.org/phpBB3/viewtopic.php?f=2&t=1872
- Replaces mode_compute=2 with Matlab's simulated annealing (Closes #814)
- Fixes bug in mode_compute=6
- Fixes bugs that lead to crashes when using user-defined mode_compute

To be done: 
- documentation when the pull request is agreed upon
- Changing of the license file

### DIFF
--- a/matlab/dynare_estimation.m
+++ b/matlab/dynare_estimation.m
@@ -91,7 +91,7 @@ else
     dynare_estimation_1(var_list,dname);
 end
 
-if options_.mode_compute && options_.analytic_derivation,
+if isnumeric(options_.mode_compute) && options_.mode_compute && options_.analytic_derivation,
     options_.analytic_derivation=analytic_derivation0;
 end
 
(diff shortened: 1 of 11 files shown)


stepan-a:
@JohannesPfeifer I agree that we have to do something for ```mode_compute=2```. But  ```saoptimset```and ```simulannealbnd``` Matlab's routines are not distributed with the Optimization toolbox, they are distributed in the Global Optimization toolbox. I do not wish to add another (optional) requirement on a Mathworks' toolbox. We could instead trigger the code contributed by Gianni (which also works with Octave) when ```mode_compute=2``` is used... And keep ```102``` for the Mathworks Global optimization routine. If you agree I do the change.

JohannesPfeifer:
Fine with me. I was just worried about backward compatibility as ```102``` was already taken for what would now move to ```2```. That was the reason I used ```2``` for the global optimization toolbox one.

stepan-a:
This was not documented in the reference manual (I only saw mentions on the forum), so it's not a problem. I do it.

JohannesPfeifer:
Yes, but I was always wondering why we had provisions for 101 and 102 in the official files. See the discussion in #814.  I simply presumed there was a purpose for these.

stepan-a:
101 and 102 are the results of very bad programing habits on our side. In this case I think @MichelJuillard  needed to use ```solvopt``` and decided to hide this in 101. After he implemented the possibility to pass a string to ```mode_compute```.

PR #878, opened 2015-04-11
Restore compatibility of recursive estimation with new data interface

JohannesPfeifer:
Use dataset_ structure instead of reconstructing data to plot. Eliminates crashes. Solves #780 for recursive estimation


stepan-a:
Manually merged. In commit 8022a181c8f7f170ef203002cdd457b570457535 I removed the excel datafile from versioning . The datafile is downloaded from www.dynare.org/Datafiles when needed.

JohannesPfeifer:
@stepan-a Perfect. I did not know how to deal with the binary Excel-file. This solves the problem. Thanks.

stepan-a:
@johannes I think we should use the same approach for other the binaries in the testsuite. But I need to give you write permissions on the server (I do not remember if you have an account on kirikou, I need to check).

JohannesPfeifer:
@stepan-a No, I only have accounts on karaba and sedna, but only the latter is working.

johannes:
Please do not write "@johannes" this will subscribe me. Thank you.

PR #880, opened 2015-04-14
GSA fixes and extensions

rattoma:
I am not sure if test routine ls2003a.mod is included in the testsuite: this is to check irf and moment calibration
still updating documentation...

### DIFF
--- a/matlab/dynare_sensitivity.m
+++ b/matlab/dynare_sensitivity.m
@@ -163,6 +163,11 @@ options_gsa = set_default_option(options_gsa,'istart_rmse',options_.presample+1)
 options_gsa = set_default_option(options_gsa,'alpha_rmse',0.001);
 options_gsa = set_default_option(options_gsa,'alpha2_rmse',1.e-5);
 
+if options_gsa.neighborhood_width,
+    options_gsa.pprior=0;
+    options_gsa.ppost=0;
+end
+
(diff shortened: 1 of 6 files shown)


stepan-a:
I agree with @marco. The issue with the size of the saved ```*.mat``` files is not so obvious and should be dealt with later. Besides I do not think that this is a problem of practical importance because the value of MaxNumberOfBytes is relatively small. This was an issue when I created this mechanism because we had much less RAM than today.

I check/fix the issue with ```ls2003a.mod``` in the testsuite.

stepan-a:
@rattoma I was not able to cherry-pick 0bd8fef2029f27e47f7668247b3ab146d1e7ed81 in 4.4... If you think that the fix should be applied to 4.4, please do a PR for the stable branch.

JohannesPfeifer:
@rattoma For those issues, that affect 4.4, could you also please add them to http://www.dynare.org/DynareWiki/KnownBugs

stepan-a:
@rattoma I already added two lines on the wiki for the commits I could cherry-pick.

rattoma:
Concerning https://github.com/DynareTeam/dynare/commit/0bd8fef2029f27e47f7668247b3ab146d1e7ed81, this is a new feature of 4.5, so no neet to cherry pick it.

stepan-a:
The test suite failed last night because of ls2003a.mod. It appears that this mod files causes a segmentation fault. I did not figure out why... I remove this mod file from the test suite and open an issue.

PR #900, opened 2015-04-20
Rename static_and_dynamic_models_differ field to static_dynamic_models_d...

JohannesPfeifer:
...iffer to fix unit tests under Linux (initval_file/ramst_initval_file.mod)

Apparently, field lengths in Octave under Linux are restricted to 31 characters. Any characters after that seem to be cut away.


JohannesPfeifer:
This seems to be a Octave bug in 3.6.2 under Linux. The problem is gone in 3.8.2.

PR #903, opened 2015-04-22
Add check for valid filename

JohannesPfeifer:
Uses the same restrictions as variable names.

### DIFF
--- a/matlab/dynare.m
+++ b/matlab/dynare.m
@@ -94,6 +94,10 @@ if ~ischar(fname)
     error('DYNARE: argument of dynare must be a text string')
 end
 
+if ~isvarname(fname)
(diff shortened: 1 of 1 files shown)


JohannesPfeifer:
@houtanb I fixed and clarified the message to stress that the filename must be a valid Matlab function name.

houtanb:
@JohannesPfeifer This commit introduces a bug. Now if we call dynare with a filename with the .mod extension, this error is reached.

Further improvements: it needs to be placed below the check for whether or not this mod file is being run from the current folder

JohannesPfeifer:
@houtanb I changed it, but could not append it to this pull request. This was more complicated than thought, because you can have periods in filenames that crash the tests before the folder check. I hope I caught all cases

PR #930, opened 2015-05-09
Make sure that no variable is named hessian

### LINKED ISSUE #667: rename hessian.m in dyn_hessian.m
In order to avoid name collision with other Matlab program/toolboxe

JohannesPfeifer:


JohannesPfeifer:
@stepan-a I removed the renaming of the hessian file and only attach cosmetic fixes to prevent name clashes within files.

PR #947, opened 2015-05-27
Change dynare_minimize_objective.m to use the correct field when sele…

JohannesPfeifer:
…cting user-defined optimizer.


stepan-a:
Manually merged.

PR #954, opened 2015-06-05
Add unit test for deterministic simulation invoking solve_one_boundar…

JohannesPfeifer:
…y.mod


houtanb:
Is the ```.mat``` file being tested? Can we avoid using it?

JohannesPfeifer:
@houtanb That is the calibration provided by the user. Unfortunately, I don't see how we can avoid it.

MichelJuillard:
We can add the program generating the .mat file

JohannesPfeifer:
I would have done so, but I don't have it as this was a user post.

MichelJuillard:
Do we need to replicate this exact example?

JohannesPfeifer:
Not necessarily, but this is the first mod-file I encountered where this particular bug appeared. If you can create an easier example that triggers the use of solve_one_boundary.m this would also work.

stepan-a:
Two remarks.

(1) We do not need to version the mat file. All the (data) binary files should be available at www.dynare.org/Datasets, and downloaded using a matlab command in the mod file (we have  already done that).

I could do that, but...

(2) I am not (yet) convinced that we should add such a model to the testsuite... I do not know the bug you are refering, but we should be able to test for it with a smaller model.

JohannesPfeifer:
@stepan-a We had a bug in the preprocessor when translating the mod-file to the ```_dynamic.m``` file when a call to ```solve_one_boundary.m``` was required. As this bug did not appear in any mod-file, it suggests that we have no file in the testsuite testing this particular capacity of the preprocessor and the involved m-files. As I said, if someone can write a simple test for this, I am all for it.

You are right that the model is complex, but the file runs in 6 seconds on my machine, so adding it might be less of a burden than coming up with a new test.

stepan-a:
Manually merged (see af29c078ad8dca3cf7754eea5518766a99006c4d and 462639103f018446830f58553677d3ca209ee726) because of the binary file.

PR #956, opened 2015-06-07
Fix bugs in dyn_mex.m under Windows

JohannesPfeifer:
Fixes string comparison and copy and paste error.

@houtanb: Line 60 (cygwin case) may also have a copy and paste error. Could it be that it should also be ```static``` there?


JohannesPfeifer:
@houtanb Did you check the cygwin case? Was it correct?

PR #982, opened 2015-07-22
TeX unit test

JohannesPfeifer:


JohannesPfeifer:
@houtanb You are right. I got rid of the eval

PR #689, opened 2014-07-15
Increase robustness of sim1.m by handling the case when A\res contains nan or inf

tholden:
sim1.m calls lin_solve to perform A\res after this change. lin_solve_robust.m is provided as an even more robust alternative which perhaps could be enabled later via an option.

See the below multi-country irreversible investment model for an example of why this helps:

```
@#define NumberOfCountries = 2
@#define Deterministic = 1

var log_kappa;
@#for Country in 1:NumberOfCountries
	var mu@{Country}, logit_l@{Country}, k@{Country}, a@{Country};
@#endfor

parameters alpha beta varrho delta theta rhoA sigmaA;

alpha = 0.3;
beta = 0.99;
varrho = 1.5;
delta = 0.025;
theta = 0.9;
rhoA = 0.95;
sigmaA = 0.05;

@#for Country in 1:NumberOfCountries
	varexo epsilonA@{Country};
@#endfor

model;

	#kappa = exp( log_kappa );
	#LEAD_kappa = exp( log_kappa(+1) );
	
	#min_A = theta ^ ( 1/alpha );
	#mean_a = log( 1 - min_A );

	@#for Country in 1:NumberOfCountries
	
		#K@{Country} 		= exp( k@{Country} );
		#LAG_K@{Country} 	= exp( k@{Country}(-1) );
		#LEAD_A@{Country} 	= min_A + exp( a@{Country}(+1) );
		#A@{Country} 		= min_A + exp( a@{Country} );
		#LAG_A@{Country} 	= min_A + exp( a@{Country}(-1) );
		#L@{Country}		= 1 / ( 1 + exp( -logit_l@{Country} ) );
		#LEAD_L@{Country}	= 1 / ( 1 + exp( -logit_l@{Country}(+1) ) );
		
		#C@{Country} = kappa ^ ( -1/varrho ) - theta / ( 1-alpha ) * ( A@{Country}*(1-L@{Country}) ) ^ ( 1-alpha );
		#LEAD_phi@{Country} = ( 1 - theta * ( LEAD_A@{Country}*(1-LEAD_L@{Country}) ) ^ ( -alpha ) ) * LEAD_kappa;
		#I@{Country} = K@{Country} - ( 1-delta ) * LAG_K@{Country};
		
	@#endfor
	
	@#for Country in 1:NumberOfCountries
	
		( a@{Country} - mean_a ) = rhoA * ( a@{Country}(-1) - mean_a ) - sigmaA * epsilonA@{Country};
[...]

PR #984, opened 2015-07-22
Sim1robust

tholden:
Cherry picked the relevant changes from the old pull request https://github.com/DynareTeam/dynare/pull/689
I misunderstood how pull requests worked initially.

### DIFF
--- a/matlab/lin_solve.m
+++ b/matlab/lin_solve.m
@@ -1,0 +1,27 @@
+function x = lin_solve( A, b )
+    if norm( b ) < sqrt( eps ) % then x = 0 is a solution
+        x = 0;
+        return
+    end
+    
+    WarningState = warning( 'off', 'all' );
+    
+    try % wrap in a try-catch block to ensure that the WarningState is always reset
+    
+        x = A\b;
+        x( ~isfinite( x ) ) = 0;
+        relres = norm( b - A * x ) / norm( b );
+        if relres > 1e-6
+            fprintf( 'WARNING : Failed to find a solution to the linear system.\n' );
+        end
+        
+    catch Error
+        
+        warning( WarningState );
+        error( Error ); %rethrow
+        
+    end
+    
+    warning( WarningState );
+    
+end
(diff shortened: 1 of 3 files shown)


JohannesPfeifer:
So if I understand the three commits correctly, in the end you just replace the left divide
```
dy = -A\res;
```
by 
```
    if norm( b ) < sqrt( eps ) % then x = 0 is a solution
        x = 0;
        return
    end
   
    WarningState = warning( 'off', 'all' );
    
    try % wrap in a try-catch block to ensure that the WarningState is always reset
        x = A\b;
        x( ~isfinite( x ) ) = 0;
        relres = norm( b - A * x ) / norm( b );
        if relres > 1e-6
            fprintf( 'WARNING : Failed to find a solution to the linear system.\n' );
        end      
    catch Error
        warning( WarningState );
        error( Error ); %rethrow
    end
    warning( WarningState );
end
```
I think that this makes a lot of sense. I was just wondering which type of errors do you have in mind that you are trying to control with the try-catch statement? My feeling is that any problems that might trigger a crash should be controlled earlier on and not within this function.

JohannesPfeifer:
@tholden A last thing. Your pull request still contains the ```lin_solve_robust.m```, but it is nowhere used now.

tholden:
The try catch is a standard "pattern" when temporarily modifying a global state. Without it, an error thrown in the inner block would leave all warnings disabled. To be honest, given that I find it helpful to disable warnings quite often, this is a block of code that I've written so many times I do it on auto-pilot, without thinking too much about what errors might be thrown in the middle (from looking at it, you could potentially get errors if b was a matrix). But since it's essentially zero cost (it is not like you can disable exception handling in MATLAB as you can in C++) it seems like good programming practice.

The lin_solve_robust.m file is still in there as an alternative which may be useful in some circumstances. Perhaps an option should be included to use it, or just trust sophisticated users to change lin_solve to lin_solve_robust if necessary. Or just delete it. It's up to you.

houtanb:
superseded by #1052

Issue #270, opened 2013-02-21
Determinant computation in marginal_density.m looks sub-optimal


sebastien-villemot:
In marginal_density.m, the determinant is computed (at three places) using the det() function. This function is known to be less precise than a Cholesky-based determinant computation when the determinant is close to zero.

Stéphane thinks that the Cholesky was used before, so we need to understand why this has changed, and therefore whether we can go back to the Cholesky.

Issue reported by Gilles Bélanger (Ministère des Finances du Québec)

JohannesPfeifer:
http://de.mathworks.com/matlabcentral/answers/41080-compute-determinant-using-cholesky-decomposition suggests that there is no point in changing this. If agreed upon, we could close this issue.

MichelJuillard:
I agree with closing the issue

JohannesPfeifer:
@MichelJuillard Sorry, I missed the loop. I amended the PR along the suggested lines

JohannesPfeifer:
@MichelJuillard We should then close this as well

PR #991, opened 2015-07-23
marginal_density.m: performance improvements

JohannesPfeifer:
Related to #270


JohannesPfeifer:
@MichelJuillard I cleaned up the part you complained about. When doing so I noticed that we only need to compute the determinant once.

PR #1000, opened 2015-07-26
Bugfixes related to collect_LaTeX_Files.m

JohannesPfeifer:
- deals with potentially unset M_.dname
- loads missing float package
- prevents crash when longnames is not used


MichelJuillard:
We should probably initialize M_.dname=M_.fname in global_initialization.m

PR #1002, opened 2015-07-28
Initialize M_.dname=M_.fname in global_initialization.m

### LINKED ISSUE #979: Make M_.dname more consistent
M_.dname is used more as an option than as a characteristic of the model file. By default, the directory name of saving output is M_.fname. When the user wants to change the name, this should be setting options_.dname, not M_.dname

JohannesPfeifer:
Closes #979


houtanb:
@JohannesPfeifer you should undo the changes you did previously. e.g. the change in ```matlab/collect_LaTeX_Files.m```. This also leads to another issue. We should just be able to revert the commit, but you included several unrelated things together in one commit. In the future, please split these into separate commits.

houtanb:
@JohannesPfeifer also you made a similar change in a commit last week. Please undo that too.

JohannesPfeifer:
What is the issue?

houtanb:
with what?

houtanb:
In 19b7f5ea7f84ab57aaade5e94c6778f122e11dc4 you added the lines:
```
+if ~isfield(M_,'dname')
+ dname = options_.dirname;
+else
+ dname = M_.dname;
+end
```
But with this change, ```dname``` is always a field of ```M_```, so this ```if``` statement can be removed. You can also remove the passing of ```options_``` to the function

houtanb:
I think you had a similar ```if``` statement in another commit last week....

JohannesPfeifer:
@houtanb Now I got it. It's done. Reverting would not have done the trick, because there were more functions affected.

houtanb:
@JohannesPfeifer Right, but it would have undone the related changes you made in that function and the git history would have been clear.

This commit (the modification of ```global_configuration.m``` would have been made with the other related changes.

Please combine all of these commits into one.

JohannesPfeifer:
@houtanb Done

PR #1006, opened 2015-07-30
4.3

mboisson:
The current parameters in the configure.ac do not work with recent version of autoreconf (using 2.69)


houtanb:
We do not update the code for old versions of dynare

PR #1008, opened 2015-07-31
Unit test checking correctness of decision rules against results from…

JohannesPfeifer:
… Dynare 4.3.2

Detects problems as the unnoticed one introduced by f26551509c4a497ee842b08a12797a7b7601db0b


houtanb:
Please change the commit message as f265515 did not introduce the bug

houtanb:
You also need to remove the ```.mat``` file from the commit.

JohannesPfeifer:
How should I remove the mat-file? I somehow need to save the previous results.

houtanb:
You can save them as a ```.csv```, load the ```.csv```, and then do the comparison

JohannesPfeifer:
Can't we put it on http://www.dynare.org/Datasets as the other binary files?

houtanb:
Ok, yes. But you need to remove the ```.mat``` files from the commit

JohannesPfeifer:
So I should also remove the one that is already in the commit history? Currently, I only do not introduce a new one.

houtanb:
Yes. Ammend the current commit removing the .mat file. Once it's in the history it ways down the git repository

JohannesPfeifer:
I removed them. How do I upload to the server?

houtanb:
I can do that. But I need to figure out how @stepan-a set this up, so may not merge it today.... Please email me the file

stepan-a:
@houtanb you just need to scp the file to ```kirikou.cepremap.org:/srv/d_kirikou/www.dynare.org/Datasets```. I have just added @JohannesPfeifer (jpfeifer) as a member of the ```dynare-web```group so he should be able to do that in the future.

PR #1017, opened 2015-08-06
Fix bugs in gsa with morris and posterior sampling

JohannesPfeifer:
- Uses defined number of observations instead of empty options_.nobs (may already solve the problem discussed in #780)
- Only saves variables that actually exist
- Adds unit test


JohannesPfeifer:
@rattoma Do you agree with the proposed changes?

houtanb:
@rattoma have you been able to test these changes?

rattoma:
Apologize. Did not have a look yet. Need to catch up since June, also because I have a lot of pushes to re-align to the rest of dynare...
thanks for reminding this.

rattoma:
@JohannesPfeifer @houtanb. Yes, I'm OK with this!

houtanb:
@rattoma thanks

PR #1019, opened 2015-08-06
Clarify unit test dependencies

### LINKED ISSUE #1018: fix test suite bugs
In email of 6/8/2015 on the dev list, @JohannesPfeifer reported dependencies of the test suite that don't make sense. fix these

JohannesPfeifer:
fs2000 is needed by other tests while fs2000_mf.mod is redundant and not called. Thus, its dependency definition can be deleted

Closes #1018


houtanb:
Why is ```fs2000_mf.mod``` redundant?

JohannesPfeifer:
It tests using a mode-file. As we have dozens of other tests loading a mode-file, it was removed from the testsuite some time ago. But we never deleted the actual file.

houtanb:
And what about ```optimizers/fs2000_102.mod``` ?

JohannesPfeifer:
It uses the mode-file created by the ```fs2000.mod```. Thus, the dependency is correctly specified.

houtanb:
Ah sorry, thought it was in ```optimizers/fs2000.mod``` didn't see that it was in fact ```estimation/fs2000.mod```

PR #1023, opened 2015-08-11
Clarify that oo_.posterior_std comes from mode computing

### LINKED ISSUE #1013: Fix saving of posterior standard deviations
See also http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=7130
When running the MCMC, the ```oo_.posterior_std``` and ```oo_.posterior_mode``` fields are based on the posterior mode, not the MCMC. But the manual says they should only be there if ```mh_replic>0``` but we always write them. 
In contrast, ```oo_.posterior_variance``` is based on the MCMC. However, this field is not documented in the manual at all. 

It is also confusing that the MCMC writes only a variance and ML/mode-finder only a standard deviation.

We must decide on a consistent treatment and fix the manual accordingly. 
My proposal is: We change the current ```oo_.posterior_std``` to ```oo_.posterior_std_at_mode``` to make the meaning explicit and let the MCMC now also write ```oo_.posterior_std```. This would be closest to what the manual states as all fields then would have the intended meaning while only adding additional fields. If need be, this would also allow going back to the 4.4 "convention" using the ```convert_dyn_45_to_44``` utility.

JohannesPfeifer:
Closes #1013

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -5975,10 +5975,19 @@ Variable set by the @code{estimation} command, if it is used with
 @end defvr
 
 @defvr {MATLAB/Octave variable} oo_.posterior_mode
-Variable set by the @code{estimation} command, if it is used with
-@code{mh_replic > 0} or @code{load_mh_file} option. Fields are of the form:
+Variable set by the @code{estimation} command during mode-finding. Fields are 
+of the form:
 @example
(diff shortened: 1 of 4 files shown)


houtanb:
Please combine these two commits

JohannesPfeifer:
@houtanb Done.

PR #1030, opened 2015-08-17
Condition test whether varobs was defined on existence of field

JohannesPfeifer:


JohannesPfeifer:
@houtanb I am not sure about this. Maybe we should think this through first. At some point, we agreed that every field of ```options``` that can exist, should also exist. This would argue for setting ```options_.varobs``` to empty in ```global_initialization``` and just drop this check for existence.

houtanb:
@JohannesPfeifer then we should make a final decision on that and implement that throughout the code for all ```options_``` fields. But, as it stands now, this commit can be simplified.

JohannesPfeifer:
@houtanb Could you then please open a ticket for this issue. I have amended the commit accordingly.

JohannesPfeifer:
@houtanb Done

PR #1037, opened 2015-08-25
Add capacities for rolling window estimation

### LINKED ISSUE #1034: Add capacities for rolling window forecasts
Currently, we only support recursive forecasting where the time window, on which the model is estimated, expands.
Necessary steps:
1. Allow ```estimation``` option ```first_obs``` to take vector like ```nobs``` (with the same consistency checks)
2. Adjust ```dynare_estimation``` to not either set ```options_.nobs = nobs(i);``` or  ```options_.first_obs=first_obs(i)``` and add check that makes them mutually exclusive.

The only problem is returning the results. Currently, ```dynare_estimation``` returns ```oo_recursive```. We can either store the rolling window estimation in the same structure or let the preprocessor assign the results to either ```oo_recursive``` or ```oo_rolling``` depending on whether ```nobs``` or ```first_obs``` is more than a scalar. In this case, the preprocessor must also do the check that either rolling or recursive estimation can be requested, but not both.

JohannesPfeifer:
Closes #1034

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -4788,12 +4788,24 @@ the file after @code{first_obs}
 Runs a recursive estimation and forecast for samples of size ranging
 of @var{INTEGER1} to @var{INTEGER2}. Option @code{forecast} must
 also be specified. The forecasts are stored in the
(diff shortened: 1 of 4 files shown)


JohannesPfeifer:
@houtanb I removed the redundant checks for empty fields

houtanb:
Great, thanks a lot.

PR #1038, opened 2015-08-25
Allow computing posterior kernel density

### LINKED ISSUE #1035: Add option to save kernel density for posterior objects
This allows for example getting the posterior density of forecasts.

It simply requires adjusting ```pm3.m``` to save the objects that can already be computed via ```posterior_moments```.
I would propose to call the option ```posterior_kernel_density``` and save it to ```options_.estimation.posterior_kernel_density.indicator=1```

JohannesPfeifer:
Closes #1035

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -5436,6 +5436,9 @@ controlled by the @code{ar} option.
 @xref{contemporaneous_correlation}. Results are stored in @code{oo_.PosteriorTheoreticalMoments}. 
 Note that the @code{nocorr}-option has no effect.
 
+@item no_posterior_kernel_density 
+Shuts off the computation of the kernel density estimator for the posterior objects (@pxref{density-field}). 
+
 @item conditional_variance_decomposition = @var{INTEGER}
 See below.
 
@@ -5814,8 +5817,10 @@ Variance of the posterior distribution
 Deciles of the distribution.
 
 @item density
-Non parametric estimate of the posterior density. First and second
-columns are respectively abscissa and ordinate coordinates.
+@anchor{density-field}
+Non parametric estimate of the posterior density following the approach outlined in @cite{Skoeld and Roberts (2003)}. First and second
+columns are respectively abscissa and ordinate coordinates. 
+
 @end table
 
 @item ESTIMATED_OBJECT
@@ -13123,6 +13128,10 @@ Schnabel, Robert B. and Elizabeth Eskow (1990): ``A new modified Cholesky algori
 Sims, Christopher A., Daniel F. Waggoner and Tao Zha (2008): ``Methods for
 inference in large multiple-equation Markov-switching models,''
 @i{Journal of Econometrics}, 146, 255--274
+
+@item
+Skoeld, Martin and Gareth O. Roberts (2003): ``Density Estimation for the
+Metropolis-Hastings Algorithm,'' @i{Scandinavian Journal of Statistics}, 30, 699--718
 
 @item
 Smets, Frank and Rafael Wouters (2003): ``An Estimated Dynamic
--- a/matlab/GetPosteriorParametersStatistics.m
+++ b/matlab/GetPosteriorParametersStatistics.m
@@ -389,6 +389,6 @@ oo.posterior_density.(type).(name) = density;
 function [post_mean,hpd_interval,post_var] = Extractoo(oo,name,type)
 hpd_interval = zeros(2,1);
 post_mean = oo.posterior_mean.(type).(name);
-hpd_interval(1) = oo.posterior_hpdinf.(type).(name); 
+hpd_interval(1) = oo.posterior_hpdinf.(type).(name);
 hpd_interval(2) = oo.posterior_hpdsup.(type).(name);
 post_var = oo.posterior_variance.(type).(name);
--- a/matlab/PosteriorIRF.m
+++ b/matlab/PosteriorIRF.m
@@ -304,12 +304,12 @@ clear STOCK_IRF_DSGE;
 for i = irf_shocks_indx
     for j = 1:nvar
         name = [deblank(M_.endo_names(IndxVariables(j),:)) '_' deblank(tit(i,:))];
-        eval(['oo_.PosteriorIRF.dsge.Mean.' name ' = MeanIRF(:,j,i);']);
-        eval(['oo_.PosteriorIRF.dsge.Median.' name ' = MedianIRF(:,j,i);']);
-        eval(['oo_.PosteriorIRF.dsge.Var.' name ' = VarIRF(:,j,i);']);
-        eval(['oo_.PosteriorIRF.dsge.deciles.' name ' = DistribIRF(:,:,j,i);']);
-        eval(['oo_.PosteriorIRF.dsge.HPDinf.' name ' = HPDIRF(:,1,j,i);']);
-        eval(['oo_.PosteriorIRF.dsge.HPDsup.' name ' = HPDIRF(:,2,j,i);']);
+        oo_.PosteriorIRF.dsge.Mean.(name) = MeanIRF(:,j,i);
+        oo_.PosteriorIRF.dsge.Median.(name) = MedianIRF(:,j,i);
+        oo_.PosteriorIRF.dsge.Var.(name) = VarIRF(:,j,i);
+        oo_.PosteriorIRF.dsge.deciles.(name) = DistribIRF(:,:,j,i);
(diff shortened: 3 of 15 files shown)


stepan-a:
@JohannesPfeifer I find the name of the option ```posterior_kernel_density```confusing. If I understand correctly, this option would trigger the estimation of the posterior density of various moments, but we still by default compute the posterior density of the parameters. I would prefer something like ```moments_posterior_density``` (kernel is not necessary I think) which would automatically trigger option ```moments_varendo```. I suppose that we will also need an option to trigger the estimation of the posterior predictive densities (I think we do not have it but I did not check).

JohannesPfeifer:
@stepan-a I renamed the option accordingly. What do you mean with posterior predictive densities? All  posterior objects are written with pm3. Thus, things like forecasts should be covered by the code.

stepan-a:
Looking the entry you wrote in the manual, I did not understand that this option applies to all the posterior computations (because it is written that ```moments_varendo``` is required). I am not sure that it is a good thing. Predictive densities (densities for the forecasts) should not be triggered with the same option (same for the posterior distribution of the latent variables, ...).

JohannesPfeifer:
@stepan-a Sorry, my mistake. ```moments_varendo``` is not required. Actually, within ```compute_moments_varendo``` we always already compute the density. The current option only applies to all objects triggered with ```options_.smoother || ~isempty(options_.filter_step_ahead) || options_.forecast``` as it is called by ```prior_posterior_statistics```. What name do you suggest? Also, I will adjust the manual.

stepan-a:
@JohannesPfeifer I still have no idea for a name. Actually I do not understand what is (should be) triggered by the option. When I ran ```tests/estimation/fs2000.mod``` I did not notice anything new in the output...

JohannesPfeifer:
In ```oo_.PointForecast``` and ```oo_.MeanForecast``` there should now be fields named ```density```.

stepan-a:
Ah... I have no such fields. I guess the ```forecast``` option is missing... When I add this option, Dynare returns the following error message (in ```pm3.m```):

```matlab
Undefined function or variable "Density".

Error in pm3 (line 169)
    oo_.(name3).density.(name) = Density(:,:,:,i);

Error in prior_posterior_statistics (line 295)
    pm3(endo_nbr,horizon,ifil(6),B,'Forecasted variables (mean)',...

Error in dynare_estimation_1 (line 487)
            prior_posterior_statistics('posterior',dataset_,dataset_info);

Error in dynare_estimation (line 92)
    dynare_estimation_1(var_list,dname);

Error in fs2000 (line 210)
oo_recursive_=dynare_estimation(var_list_);

Error in dynare (line 223)
evalin('base',fname) ;
```

JohannesPfeifer:
My mistake. Please try again with the updated files including the updated fs2000.mod

stepan-a:
I will check. By the way, it would be much easier to track your branches if you avoided forced updates (it is considered as bad practice to rewrite history in Git). When you fix something just add new commits.

JohannesPfeifer:
@houtanb wanted me to amend previous commits instead of adding new commits, but this means I need to force the update to Github.

houtanb:
@stepan-a it's better for the final git history for this to be overwritten. We don't want to see all the work that went into the final version of a commit, just the commit itself

houtanb:
also, this is the way they do it in other projects where they accept commits via pull requests

stepan-a:
@houtanb Why not... Even if I disagree with the first statement, I believe that we learn a lot from errors and that this kind of noise convey information. But, given my limited knowledge of Git, this ends up in more work on my side. How do you manage to review (before merging) the new commits then (without locally delete the branch)?

stepan-a:
@JohannesPfeifer Your last modification fixes the error... But I see no ```density``` field in ```oo_.PointForecast``` or ```oo_.MeanForecast```. I suppose this is not intended.

JohannesPfeifer:
No. I only now realize that @houtanb created the option under ```options_``` while I wanted it under ```options_.estimation```. That explains why the code is never triggered.

JohannesPfeifer:
@stepan-a I would suggest you test the code by setting ```options_.estimation.moments_posterior_density.indicator=1``` manually and we fix this when we decide on the final name of the option.

stepan-a:
Agreed.

stepan-a:
Okay. This fixes the problem. But I am a bit surprised by the results. If I look at the predictive densities for y, they are almost equal at each step (h=1, 2, ..., 8) for PointForecast and MeanForecast (I still do not quite understand the names of these fields).

I still feel uneasy with this option (whatever the name)... I suspect that the difficulty we face in choosing a good name comes from the fact that the option itself is debatable. I have the feeling that we should always compute these densities (as for the posterior moments) and that we should instead propose an option to turn off the computation of all the densities (posterior densities for the parameters, posterior densities for the moments, ...).

JohannesPfeifer:
@stepan-a Regarding the names, I also had a hard time. But I documented them some time ago in https://sites.google.com/site/pfeiferecon/Pfeifer_2014_Dynare_graphs.pdf?attredirects=0 :
```The mean forecasts only take the parameter uncertainty into account, but omit the uncertainty about future shocks. Future shocks are averaged out/set to 0. In contrast to the mean forecasts, the points forecast not only take the parameter uncertainty into account, but also take into consideration the uncertainty about future shocks.```

For y this behavior is to be expected. When you look at the forecast graphs, the deciles are also flat. I think the reason is that we estimate a bunch of parameters that determine the level of y. The forecast essentially says that y fluctuates a bit around this steady state value. Due to the graph reflecting mostly parameter uncertainty, there is no mean reversion in levels (only in deviations from steady state).

The reasons I did not generally enable the computation of densities are backward compatibility and that it might add a lot to computation time for larger models

houtanb:
@stepan-a if you're on a branch that @JohannesPfeifer has pushed and he has done a force update to this branch, you just do a ```git fetch``` from his repository and a ```git reset --hard``` to ```johannes/<<branch_name>>>```. You do not keep the previous commits because they are now invalid. If during the review process you want to see the incremental changes that @JohannesPfeifer makes, then he can not do force updates, creating a commit for every change he makes. Then, once you everything is ok, @JohannesPfeifer can do a ```git rebase -i origin/SHA```, squashing the commits showing the incremental work into one commit representing the logical change that was made. The point of git is to show the logical changes that were made, not to show the work that went into making a commit.

stepan-a:
Yes it must be that there is no mean reversion in this case... But when we take into account the uncertainty about future shocks the variance of the forecast error should change and this has to be reflected in the shape of the predictive density. In the case of an AR(1), for instance, we know that this variance grows with h.

I believe that the estimation of the density is far less time consuming than the computation of the deciles (where we need to sort the draws) and HPD intervals, but I will check.

stepan-a:
@houtanb Yes this is what I do...

JohannesPfeifer:
@stepan-a This is exactly what we observe for m, where the steady state does not move with parameters. The mean forecasts that do not take future shocks into account show mean reversion. In contrast, the Point Forecasts show rather constant uncertainty. You are right that for an AR1 the variance should grow with h and then converge to the unconditional variance. But this is only the case if you know the initial starting point to be at the mean. If you are far away from the mean and the mean reversion is strong, there will be no increase in forecast error variance as the mean reversion is going to dominate the effect of new shocks. Remember, we are not looking at classical forecasts, but Bayesian ones. We show the distribution over all draws, not a confidence interval around a particular forecast.

houtanb:
@stepan-a then I didn't understand your question above or your problem with this way of using pull requests

JohannesPfeifer:
@stepan-a How about computing the density always by default and then naming the option to shut it off ```no_posterior_kernel_density```. It would then suppress all output of densities, the new ones from ```pm3``` and the old ones from ```compute_moments_varendo```

stepan-a:
@JohannesPfeifer Yes, this is precisely what I was proposing [yesterday](https://github.com/DynareTeam/dynare/pull/1038#issuecomment-136856910). I still wonder if this option should also turn off the estimation of the posterior density for the parameters. This would be more consistent with the name of the option. Also, I would like to check that the computation of these densities does not slow down too much the execution of mod files.

JohannesPfeifer:
@stepan-a That's why I proposed this specific name for the option. For fs2000.mod, it adds about 7 seconds on my machine

JohannesPfeifer:
@stepan-a Following our discussion at the Dynare conference, I implemented the desired default behavior: the kernel density is now always computed, but ```no_posterior_kernel_density``` can be used to suppress it.

stepan-a:
Thanks @JohannesPfeifer.

PR #1049, opened 2015-08-31
Add unit test for Ramsey policy that checks correctness with analytic…

JohannesPfeifer:
…al results


JohannesPfeifer:
@houtanb I fixed the naming.

PR #1052, opened 2015-09-06
Integrate robust linear solver for stack_solve_algo=0

JohannesPfeifer:
Supersedes #984 

To be done: add interface option for ```robust_lin_solve``` to set ```options_.simul.robust_lin_solve=1```

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -2879,6 +2879,9 @@ add new variables one by one.
 Determines the maximum number of iterations used in the non-linear solver. The
 default value of @code{maxit} is 50. 
 
+@item robust_lin_solve
+Triggers the use of a robust linear solver for the default @code{solve_algo=4}. 
+
 @item solve_algo = @var{INTEGER}
 @anchor{solve_algo}
 Determines the non-linear solver to use. Possible values for the option are:
--- a/matlab/global_initialization.m
+++ b/matlab/global_initialization.m
@@ -63,6 +63,7 @@ options_.minimal_workspace = 0;
 options_.dp.maxit = 3000;
 options_.steady.maxit = 50;
 options_.simul.maxit = 50;
+options_.simul.robust_lin_solve = 0;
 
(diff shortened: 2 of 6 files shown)


JohannesPfeifer:
Sidenote: I don't know what changed in the unstable version, but in case of multiple solutions, the unstable version returns a different solution than the stable version. This pull request seems to restore the previous behavior.

houtanb:
@JohannesPfeifer should we close #984?

Once this is merged, please create an issue for the interface change

JohannesPfeifer:
Yes, this closes #984

MichelJuillard:
@JohannesPfeifer, 
1. what is the reference for Tom Holden robust linear solver?
2. Our experience is that bigcstab and gemres rarely work for this type of models
3. We need to think how to organize the multiple options available to solve the perfect foresight model

JohannesPfeifer:
@MichelJuillard I don't know the reference and don't have experiences with the ```robust_lin_solve```. Maybe @tholden can help out

tholden:
Sorry for the delayed reply. I created these files when having some problems with perfect foresight solutions to a model with multiple occasionally binding constraints. lin_solve_robust takes a super paranoid, try everything approach, as a final desperate attempt for very stubborn models, and the pseudo-inverse (the last thing it tries) really ought to always work, though it will be slow (hence trying it last). @jonswarbrick ought to have the latest mod file for the model which was originally giving problems (Jonathan: this is the perfect foresight version of our occasionally binding financial frictions paper).

JohannesPfeifer:
@stepan-a I talked to Tom and would like to add this as an option the user can try if everything else fails. The idea is to sequentially test approaches to "inverting" an ill-conditioned A in the hope that one of them works. If none of them does, use the pseudo-inverse as a last resort. This option of couse won't be quick and clearly is not a default choice, but might be useful in pathological cases. Therefore, I don't see why we should not support it.

PR #1062, opened 2015-09-22
Allows to compute the extended path and the conditional forecasts wit…

FerhatMihoubi:
…h bytecode


JohannesPfeifer:
@FerhatMihoubi Do you also have a mod-file to test these functions? I don't think that any of these features are already triggered in the testsuite.

stepan-a:
@FerhatMihoubi Did you receive the email I sent you last week with the error messages returned by your test for this PR?

PR #1075, opened 2015-10-11
Condition check for stochastic singularity on maximum of non-missing …

### LINKED ISSUE #1071: initial estimation checks
The first checks in `initial_estimation_checks.m` are not robust to the case where for example we use mixed frequency data.
Assume we have annual data in time interval 1995-2000 and quarterly data in 2001-2015, I would define the annual variable and the quarterly variable as BOTH observed, but with missing values mutually exclusive, i.e. in 1995-2000 Q data will be NaN, while in 2001-2015 the annual data will have been set to NaN.
This means it is not sufficient to check the # of observables declared, but we need to check the max number of non missing data values in each time point, and confront this to the number of non zero shocks declared.

JohannesPfeifer:
…observations

closes #1071

### DIFF
--- a/matlab/initial_estimation_checks.m
+++ b/matlab/initial_estimation_checks.m
@@ -36,11 +36,15 @@ function DynareResults = initial_estimation_checks(objective_function,xparam1,Dy
 % You should have received a copy of the GNU General Public License
 % along with Dynare.  If not, see <http://www.gnu.org/licenses/>.
 
-if DynareDataset.vobs>Model.exo_nbr+EstimatedParameters.nvn
+%get maximum number of simultaneously observed variables for stochastic
+%singularity check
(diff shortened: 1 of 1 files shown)


MichelJuillard:
@JohannesPfeifer I recently encountered the following case: because data availability, one changes observed variable in the middle of the sample. So, there is one more observed variable that there is shocks, but there is always one of the observed variables that has missing observation. Currently, Dynare detects falsely stochastic singularity. Would this condition fix my issue?

rattoma:
@MichelJuillard . Indeed I think this should tackle exactly this kind of problem. @JohannesPfeifer am I right?

JohannesPfeifer:
@MichelJuillard Yes, exactly.

PR #1091, opened 2015-10-15
Better handling of NPD MH proposal covariance

nerdcha:
Hi there, thanks a bunch for all your work on Dynare! Do you accept pull requests?

This is a tweak I always make to my local copy. Obviously an NPD Hessian is less than ideal, but with this change you can run the MCMC for a few thousand draws, which almost always makes it obvious where the problem is. The `mode_check` feature is a powerful tool, but doesn't catch everything.

### DIFF
--- a/matlab/metropolis_hastings_initialization.m
+++ b/matlab/metropolis_hastings_initialization.m
@@ -80,7 +80,13 @@ nblck = options_.mh_nblck;
 nruns = ones(nblck,1)*options_.mh_replic;
 npar  = length(xparam1);
 MAX_nruns = ceil(options_.MaxNumberOfBytes/(npar+2)/8);
-d = chol(vv);
+[ d, notPosDef ] = chol(vv);
+if notPosDef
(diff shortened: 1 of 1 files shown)


stepan-a:
@nerdcha We do accept pull requests. We could add this as an option. But I wonder if there is an advantage compared to running the MCMC starting with a diagonal covariance matrix (unit matrix or prior variances on the diagonal). Usually, the Hessian matrix is wrong because the optimizer did not find a maximum. In this case, I wonder if your matrix brings more useful information (w.r.t. the convergence of the MCMC) than a diagonal matrix based on the prior distribution.  Can you elaborate on what kind of problems you find with the MCMC?

nerdcha:
Now that you mention it, I hadn't noticed that you've added more options via `mcmc_jumping_covariance`. I was thinking of cases where you don't necessarily expect convergence, but where a few thousand draws makes it obvious where the problem with your model is. So the `prior_variance` or `identity_matrix` options would work just as well. Keep up the good work, eh!

PR #1095, opened 2015-10-16
Remove fsolve dependency from JFVV model

nerdcha:
Thanks @JohannesPfeifer for including this mod file in the examples; a big time-saver! These changes allow the model to be solved without the Optimization Toolbox, which might make it a bit more accessible.

### DIFF
--- a/examples/NK_baseline.mod
+++ b/examples/NK_baseline.mod
@@ -86,7 +86,7 @@ predetermined_variables k;
 
 parameters h            //consumption habits
            betta        //discount factor
-           gammma1      //capital utilization, linear term
+           //gammma1      //capital utilization, linear term
            gammma2      //capital utilization, quadratic term
            delta        //depreciation rate
            kappa        //capital adjustment costs parameter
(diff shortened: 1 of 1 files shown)


stepan-a:
@nerdcha The point of this example is to show how to use a ```steadystate``` file with a Newton inside for solving a concentrated version of the static model. I am not sure that your PR helps a lot (Also I do not know the details of this model but the basic(s) NK model I have in mind do have a closed form solution for the steady state).

That said, it is quite true that the dependency of this example on the Matlab's Optimization Toolbox is annoying. Everybody should be allowed to run this example. Dynare provides other solvers (Newton with line search and trust regions  algorithms, ```csolve``` routine developed by Chris Sims, ...). So if the issue is about removing the dependency w.r.t. Matlab's Optimization Toolbox, I would replace ```fsolve```by another routine in the ```steadystate``` file. The available routines can be seen in the ```dynare_solve``` routine (where you can figure out how to use them). It would be even nicer to call ```dynare_solve```.

nerdcha:
Oh right, I see. I like this model because it's basically Smets-Wouters, but they've done the hard work of figuring out the nonlinear equations for Calvo pricing etc. It has a single equation at the core of the steady state which needs a numerical root finder.

I'll take another swing at removing `fsolve`, using `dynare_solve` instead.

MichelJuillard:
@nerdcha, if it is a single equation that needs to be solve, use fzero(). It is in Matlab base, not in a toolbox and more efficient than dynare_solve when there is only one equation.

JohannesPfeifer:
@nerdcha I did the replacement in #1096

nerdcha:
Thanks, that's great!
>
>

Issue #992, opened 2015-07-23
Fix implementation of lik_init==5 part of dsge_likelihood


JohannesPfeifer:
```dsge_likelihood.m``` still has the part

```
  case options_.lik_init == 5            % Old diffuse Kalman filter only for the non stationary variables
    [eigenvect, eigenv] = eig(T);
    eigenv = diag(eigenv);
    nstable = length(find(abs(abs(eigenv)-1) > 1e-7));
    unstable = find(abs(abs(eigenv)-1) < 1e-7);
    V = eigenvect(:,unstable);
    indx_unstable = find(sum(abs(V),2)>1e-5);
    stable = find(sum(abs(V),2)<1e-5);
    nunit = length(eigenv) - nstable;
    Pstar = options_.Harvey_scale_factor*eye(np);
    if kalman_algo ~= 2
        kalman_algo = 1;
    end
    R_tmp = R(stable, :);
    T_tmp = T(stable,stable);
    if DynareOptions.lyapunov_fp == 1
        Pstar_tmp = lyapunov_symm(T_tmp,R_tmp*Q*R_tmp',DynareOptions.lyapunov_fixed_point_tol,DynareOptions.qz_criterium,DynareOptions.lyapunov_complex_threshold, 3, [], DynareOptions.debug);
    elseif DynareOptions.lyapunov_db == 1
        Pstar_tmp = disclyap_fast(T_tmp,R_tmp*Q*R_tmp',DynareOptions.lyapunov_doubling_tol);
    elseif DynareOptions.lyapunov_srs == 1
        Pstar_tmp = lyapunov_symm(T_tmp,Q,DynareOptions.lyapunov_fixed_point_tol,DynareOptions.qz_criterium,DynareOptions.lyapunov_complex_threshold, 4, R_tmp, DynareOptions.debug);
    else
        Pstar_tmp = lyapunov_symm(T_tmp,R_tmp*Q*R_tmp',DynareOptions.qz_criterium,DynareOptions.qz_criterium,DynareOptions.lyapunov_complex_threshold, [], [], DynareOptions.debug);
    end    
    Pstar(stable, stable) = Pstar_tmp;
    Pinf  = [];
```
that can never be reached. ```options_``` does not even exist in this function. I propose deleting this part.

stepan-a:
I fixed the bug relaed to ```options_``` in commit d125c2effc2b7c1d103d96173ddc4b59c7ee5749. I am not sure I want to delete this part of the code.

[...]

PR #1097, opened 2015-10-18
Document lik_init=5 and add unit test for lik_init

JohannesPfeifer:
Related to #992

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -4898,13 +4898,19 @@ variables
 
 @item 2
 For nonstationary models: a wide prior is used with an initial matrix
-of variance of the error of forecast diagonal with 10 on the diagonal
+of variance of the error of forecast diagonal with 10 on the diagonal 
+(follows the suggestion of @cite{Harvey and Phillips(1979)})
 
 @item 3
 For nonstationary models: use a diffuse filter (use rather the @code{diffuse_filter} option)
 
 @item 4
 The filter is initialized with the fixed point of the Riccati equation
+
+@item 5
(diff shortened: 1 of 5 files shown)


JohannesPfeifer:
@stepan-a The implementation of ```lik_init=5``` is still not complete. See the crashes in the unit tests

stepan-a:
Thanks, I try to fix this before merging (do not want to add a new failing test for a hidden feature).

stepan-a:
I fixed the issue with lik_init equal to 5 (I pushed the fix in a personal branch [here](https://gitlab.ithaca.fr/Dynare/dynare/commits/lik_init_5) where I also rewrote the tests) but I found another issue with ```lik_init==3``` with the estimation in levels of fs2000...

JohannesPfeifer:
Which issue do you encounter?

stepan-a:
```schur_statespace_transformation``` routine returns NaNs for ```Pstar```... But it's most likely an issue with my test.

stepan-a:
@JohannesPfeifer I rebased my branch [here](https://gitlab.com/stepan-a-dynare/dynare/commits/lik_init_5) on master. All remaining issues are fixed. Did you intend to compare likelihood values returned for different initializations? These values, in some cases, are very different, and this is expected. So I do not see the point in such a comparison (I may have misunderstood your idea)... But I suppose that these discrepancies should narrow when the sample size is increased. In the current state, at least, we test that all the options for initialization are working.

JohannesPfeifer:
@stepan-a Could it be that this change needs to be ported to DsgeSmoother.m?

PR #1102, opened 2015-11-07
Julia bugfix in assert and assign params in __modfile__.jl

spencerlyon2:
I found a bug in an `@assert` call in the first generated `dynamic!` method in the Julia code (`size` always returns a tuple, and we were comparing to an int. Dispatch restricts these arguments to all be vectors so `length` works here)

Also added support for filling in the `model.params` `Vector`.

I realize that the Julia support is still preliminary, but I'm happy to pitch in as I have time/need to push things forward.


stepan-a:
Manually merged.

PR #1104, opened 2015-11-16
Only accept steady state when fsolve retuns with normal return code 1

JohannesPfeifer:
Otherwise, termination due to tolerance criteria will be accepted as correct steady state

### DIFF
--- a/matlab/dynare_solve.m
+++ b/matlab/dynare_solve.m
@@ -113,8 +113,16 @@ if options.solve_algo == 0
         end;
     end
 
-    if exitval > 0
+    if exitval == 1
         info = 0;
+    elseif exitval > 1
+        M=evalin('base','M_'); %get variable names from workspace
+        resid = evaluate_static_model(x,varargin{:},M,options);
+        if max(abs(resid)) > 1e-6
+            info = 1;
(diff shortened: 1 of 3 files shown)


MichelJuillard:
I guess that you mean accepting when x doesn't move enough anymore. I'm not sure that it is as simple. In a badly normalized system x not moving enough may indicate the solution even if the residuals remain relatively large. We should at least display enough information so that the user can make his/her own judgment on the quality of that possible solution.   
We probably also need an option for the user accepting a solution when x doesn't move anymore.

JohannesPfeifer:
@MichelJuillard I have updated the commit. 

This issue points to a more general problem. If we accept exitcodes like
```
2: Change in x was smaller than the specified tolerance.
3 : Change in the residual was smaller than the specified tolerance.
4: Magnitude of search direction was smaller than the specified tolerance.
```
as a detected steady state, we are potentially computing the decision rules at a point that is not the steady state without warning the user. Thus, for any abnormal exit codes, we need to check the residuals. This needs to be implemented for the other optimizers as well.

MichelJuillard:
I entirely agree that the user should be warned

PR #1115, opened 2015-12-10
Only test non-zero measurement error covariance entries for positive …

JohannesPfeifer:
…definiteness

Otherwise, not having measurement error on one variable is not allowed during estimation

### DIFF
--- a/matlab/dsge_likelihood.m
+++ b/matlab/dsge_likelihood.m
@@ -210,7 +210,7 @@ H = Model.H;
 
 % Test if Q is positive definite.
 if ~issquare(Q) || EstimatedParameters.ncx || isfield(EstimatedParameters,'calibrated_covariances')
-    [Q_is_positive_definite, penalty] = ispd(Q);
+    [Q_is_positive_definite, penalty] = ispd(Q(EstimatedParameters.Sigma_e_entries_to_check_for_positive_definiteness,EstimatedParameters.Sigma_e_entries_to_check_for_positive_definiteness));
     if ~Q_is_positive_definite
         fval = objective_function_penalty_base+penalty;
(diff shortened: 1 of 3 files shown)


JohannesPfeifer:
@MichelJuillard @stepan-a My intution is that the same applies to ```M_.Sigma_e``` as otherwise we cannot have a 0 variance exogenous shock in the model.

JohannesPfeifer:
@MichelJuillard Following our discussion, I have moved the respective parts to ```dynare_estimation_init.m```.

stepan-a:
@JohannesPfeifer When I run the testsuite on your branch I obtain a lot of errors on estimation related tests (see [here](https://gitlab.com/snippets/11564)).

In ```estimation/fs2000``` matlab complains about an unknown field in ```estim_params_```called ```nvn_observable_correspondence```... I do not know where it is supposed to come from. Any idea ?

JohannesPfeifer:
@stepan-a My mistake. I updated the codes. Could you please try again?

stepan-a:
Okay. I'll do that... But I am waiting for the results of the testsuite on PR [#1116](https://github.com/DynareTeam/dynare/pull/1116/commits).

stepan-a:
The testsuite was failing on ```fs2000/fs2000_calib.mod```, ```fs2000/fs2000_calib_dseries.mod``` and 
```kalman_filter_smoother/fs2000_smoother_only.mod``` because ```estim_params_.var_exo```is not defined when running a smoother on a calibrated model. This issue is fixed in 846a391eb11321a5192cb552b9fa0416bdae8c5c.

stepan-a:
Manually merged.

PR #1118, opened 2015-12-17
Fix two unit tests

JohannesPfeifer:


stepan-a:
Thanks Johannes. I was about to push the same fix for the comparison of marginal densities... But I still have some doubts about this test. I think that it would be better to do the comparison when we know that the two estimates have to return the same value (in a linear gaussian model for instance).

stepan-a:
@JohannesPfeifer Sorry, looking at the testsuite results, I have just realized that I clicked on the wrong link this afternoon (from my cell phone) and closed your PR (without merging) unintentionally...

PR #1132, opened 2016-03-11
Fix incompatibility of extended_path.m with parfor (thanks to Tom Hol…

### LINKED ISSUE #1127: Minor bug in extended_path.m
A parfor loop contains the line (211):
`[results{k}(:,t), info_convergence] = extended_path_core(ep.periods,endo_nbr,exo_nbr,positive_var_indx, ...`
Later, `info_convergence` is used outside of the parfor loop (line 236).
MLint detects this as a critical bug, though I expect it hadn't been hit as people had run with verbosity off.
The fix is to make `info_convergence` a reduction variable. (Outside the loop `info_convergence = 1`, inside the loop `[results{k}(:,t), tmp_info_convergence] = extended_path_core(ep.periods,endo_nbr,exo_nbr,positive_var_indx, ...` and then `info_convergence = min( tmp_info_convergence, info_convergence );`.)

JohannesPfeifer:
…den)

Closes #1127

### DIFF
--- a/matlab/ep/extended_path.m
+++ b/matlab/ep/extended_path.m
@@ -14,7 +14,7 @@ function [ts,results] = extended_path(initial_conditions,sample_size)
 %
 % SPECIAL REQUIREMENTS
 
-% Copyright (C) 2009-2015 Dynare Team
+% Copyright (C) 2009-2016 Dynare Team
 %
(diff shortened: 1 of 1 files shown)


JohannesPfeifer:
Will be dealt with in bigger change to function

PR #1135, opened 2016-03-15
Move resetting and saving of initial seeds before first random number…

### LINKED ISSUE #1048: Eliminate remaining randomness from estimation routines
Related to http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=7243&p=21224#p21224 
There still seem to be elements where we do not make sure that estimation becomes "deterministic". For example, we only reset the seed in ```metropolis_hastings_initialization``` after we have initialized the chains via 
```
candidate = rand_multivariate_normal( transpose(xparam1), d * options_.mh_init_scale, npar);
```
thereby introducing an element of randomness. This seems to be particularly relevant if optimizers are used that rely on random numbers.

JohannesPfeifer:
…s are used in estimation

Otherwise, the starting values of the chain(s) will depend on the state of the RNG.

Closes #1048

### DIFF
--- a/matlab/metropolis_hastings_initialization.m
+++ b/matlab/metropolis_hastings_initialization.m
@@ -37,7 +37,7 @@ function [ ix2, ilogpo2, ModelName, MetropolisFolder, fblck, fline, npar, nblck,
 % SPECIAL REQUIREMENTS
 %   None.
 
-% Copyright (C) 2006-2015 Dynare Team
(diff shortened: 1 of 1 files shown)


JohannesPfeifer:
This does not solve the issue as the recorded seed is used in e.g. ```random_walk_metropolis_hastings_core``` and therefore needs to be saved after drawing the starting values.

PR #1139, opened 2016-03-17
Correct manual on Ramsey

### LINKED ISSUE #944: Fix description of Ramsey timeless perspective in manual
Michel writes: 

```
I don't agree with the part on timeless perspective. Setting the initial
multiplier to its steady state value is not timeless
perspective. "Setting the initial Lagrange multipliers as if optimal
policy was started a long time ago" would take into consideration the
value of the initial state variables.
```

JohannesPfeifer:
Closes #944

### DIFF
--- a/doc/bvar-a-la-sims.tex
+++ b/doc/bvar-a-la-sims.tex
@@ -11,7 +11,7 @@
 
 \begin{document}
 
-\title{BVAR models ``\`a la Sims'' in Dynare\thanks{Copyright \copyright~2007--2011 S\'ebastien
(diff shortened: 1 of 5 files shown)


stepan-a:
@MichelJuillard Can you check if we can merge this ?

MichelJuillard:
I have a problem with the expression "the Ramsey policy is computed with the Lagrange multiplier being at the steady state" This is unclear because a policy is a rule--a function of state variables--not a trajectory.
I will think about a better formulation.

JohannesPfeifer:
@MichelJuillard I meant ```is computed by approximating the equilibrium system around the perturbation point where the Lagrange multipliers are at their steady state```.

JohannesPfeifer:
@MichelJuillard I did another iteration along my suggestion above.

PR #1140, opened 2016-03-17
Add unit tests for perfect foresight to detect issues with histval

JohannesPfeifer:


JohannesPfeifer:
@stepan-a I adjusted the unit test correspondingly.

JohannesPfeifer:
@stepan-a I also added a test that checks the issue with homotopy changing the historical values.

PR #1174, opened 2016-04-20
Slice sampler

rattoma:
This branch adds the slice sampler by factorizing at the same time posterior sampling iteration.
This means all samplers [RW metropolis, independent metropolis and slice] use the same routines to run:
`posterior_sampler`
`posterior_sampler_core`
while 
`posterior_sampler_iteration` triggers the individual sampler at each iteration.

@JohannesPfeifer in principle this could be done also for TaRB, but I did not added yet TaRB steps into `posterior_sampler_iteration`, so I still call TaRB from `posterior_sampler`.

Under this setting, any new posterior sampler (including slice) is already automatically parallelized.

If things are OK, the following routines become isolated and no longer used:
`metropolis_hastings_initialization`
`random_walk_metropolis_hastings.m`
`random_walk_metropolis_hastings_core.m`
`independent_metropolis_hastings.m`
`independent_metropolis_hastings_core.m`

The new routines
`posterior_sampler_initialization.m`
etc
are already merged merged with recent new commits by @JohannesPfeifer in particular about `mh_recover`.
I was unsure whether it would have been better just a renaming of the parent routines 

`metropolis_hastings_initialization`
`random_walk_metropolis_hastings.m`
`random_walk_metropolis_hastings_core.m`

from which I started to develop

`posterior_sampler_initialization`
`posterior_sampler.m`
`posterior_sampler_core.m`


matlab/McMCDiagnostics.m
```diff
@@ -69,6 +69,46 @@
...
+TotalNumberOfMhFiles = sum(record.MhDraws(:,2));
+TotalNumberOfMhDraws = sum(record.MhDraws(:,1));
+FirstMhFile = record.KeepedDraws.FirstMhFile;
+NumberOfDraws = TotalNumberOfMhDraws-floor(options_.mh_drop*TotalNumberOfMhDraws);
+
+param_name=[];
+for jj=1:npar   
+    param_name = strvcat(param_name,get_the_name(jj,options_.TeX,M_,estim_params_,options_));
+end
+fprintf('\nMCMC Inefficiency factors per block.\n');
```
JohannesPfeifer: @rattoma Could you please replace the manual creation of a table to be displayed by a call similar to
```
    my_title='MATRIX OF COVARIANCE OF EXOGENOUS SHOCKS';
    labels = deblank(M_.exo_names);
    headers = char('Variables',labels);
    lh = size(labels,2)+2;
    dyntable(my_title,headers,labels,M_.Sigma_e,lh,10,6);
    if options_.TeX
        labels = deblank(M_.exo_names_tex);
        headers = char('Variables',labels);
        lh = size(labels,2)+2;
        dyn_latex_table(M_,my_title,'covar_ex_shocks',headers,labels,M_.Sigma_e,lh,10,6);
    end
```
That would in particular automatically create LaTeX output

matlab/missing/stats/autocorr.m
```diff
@@ -0,0 +1,40 @@
...
+% along with Dynare.  If not, see <http://www.gnu.org/licenses/>.
+
+
+y=y(:);
+acf = NaN(ar+1,1);
+acf(1)=1;
+m = mean(y);
+sd = std(y,1);
+for i=1:ar,
+    acf(i+1) = (y(i+1:end)-m)'*(y(1:end-i)-m)./((size(y,1))*sd^2);
```
JohannesPfeifer: @rattoma I don't think it is a good idea to rely on the theoretical definition here that assumes the two series have the same standard deviation. By dividing by ```sd^2``` instead of ```std(y(i+1:end))``` and ```std(y(1:end-i))``` you can get correlations bigger than 1 in absolute value. Some time ago we decided for that reason to always rely on the finite sample definition rather than the asymptotic one.

matlab/dynare_estimation_1.m
```diff
@@ -366,6 +366,8 @@ function dynare_estimation_1(var_list_,dname)
     xparam1 = bayestopt_.p5;
     idNaN = isnan(xparam1);
     xparam1(idNaN) = bayestopt_.p1(idNaN);
+    outside_bound_pars=find(xparam1 < bounds.lb | xparam1 > bounds.ub);
+    xparam1(outside_bound_pars) = bayestopt_.p1(outside_bound_pars);
```
JohannesPfeifer: @rattoma I do not understand the underlying issue. How can it be that the mode is outside of the bounds? If this happens, haven't we done a mistake earlier on in computing the mode?

matlab/missing/stats/autocorr.m
rattoma: @JohannesPfeifer: concerning the decision on finite sample definition, are you referring to, e.g. autocovariances computed by DYNARE for stochastic simulations?
Concerning this autocorr routine, I think it would be OK to use finite sample as well. 

However, there might be an issue regarding its placement as 'missing', since currently it should reproduce the behavior of autocorr of econometrics toolbox [I moved the routine accordingly in a subsequent commit] which as far as I can see uses the asymptotic definition.
Then it would be better to rename it as dyn_autocorr and place it in the root matlab directory?

matlab/dynare_estimation_1.m
rattoma: Indeed this happened to me in some test and needed to trap it. 
Note that with slice the default is not to optimize, so there we would have the prior mode.
If I remember well, that error could happen for priors like parameter `rho` in fs2000, with the vertical asymptote at a bound combined with a prior truncation?
Or in general a prior truncation that excludes the prior mode?

matlab/McMCDiagnostics.m
rattoma: @JohannesPfeifer thanks for pointing this out. I will look into this

matlab/missing/stats/autocorr.m
JohannesPfeifer: @rattoma I see. Matlab's autocorr uses the particular definition of the autocorrelation function of 
```Box, G. E. P., G. M. Jenkins, and G. C. Reinsel. Time Series Analysis: Forecasting and Control. 3rd ed. Englewood Cliffs, NJ: Prentice Hall, 1994.```
I think the use of ```size(y,1)``` as a divisor introduces a deliberate downward bias to prevent it being bigger than 1. At a minimum, we need to document this. I still would prefer to use the actual in-sample correlation as when computing simulated moments.
Regarding your second point, calling it ```dyn_autocorr ``` would be better, because it does not really mirror the function of the econometrics toolbox, which does plotting and provides confidence bounds.

File: matlab/dynare_estimation_1.m (hunk 3 above)
Comment (JohannesPfeifer) [inline matlab/dynare_estimation_1.m]: @rattoma But your test is not whether we are at the bound, but whether we are outside of it. This should never happen. When we compute the prior mode, it must be restricted to be within the bounds or we are making a mistake. Therefore, we should rather adjust ```compute_prior_mode``` than ```dynare_estimation_1```

File: matlab/missing/stats/autocorr.m (hunk 2 above)
Comment (rattoma) [inline matlab/missing/stats/autocorr.m]: @JohannesPfeifer OK for `dyn_autocorr`. Would you suggest to redefine `acf` as follows, then?
```
sd1 = std(y(i+1:end),1);
sd2 = std(y(1:end-i),1);
acf(i+1) = (y(i+1:end)-m)'*(y(1:end-i)-m)./((size(y,1)-i)*sd1*sd2);
```

Comment (JohannesPfeifer) [inline matlab/missing/stats/autocorr.m]: @rattoma Exactly.

**Comment by JohannesPfeifer** [comment]
@rattoma Is there a particular reason you treat TaRB differently? It should follow exactly the same logic as the other cases, I guess.

**Comment by rattoma** [comment]
the reason is just I did not have time, after looking into the TaRB code, to analyze it in detail and snip out properly the individual sampler iteration step.
I should have contacted you at that time.

File: matlab/missing/stats/autocorr.m (hunk 2 above)
Comment (rattoma) [inline matlab/missing/stats/autocorr.m]: By the way, I think that MATLAB's autocorr prevents acf >1 because:
`sd = std(y,1);` [i.e. sum of squares divided by size y] and the term `size(y,1)` exactly removes the denominator from `sd^2`
then the numerator sums  `size(y,1)-i` terms while denominator sums `size(y,1)` terms.
With this form, autocovariances and autocorrelations are just transformed into each other by the common factor `sd^2`, at any `i`. I am not sure whether this is a deliberate downward bias. 
In any case the finite sample definition has also a clear meaning, so we can go with that.

**Comment by JohannesPfeifer** [comment]
@rattoma Ok. It should be straighforward. I will try to look into it. 
@stepan-a Is the plan to include this in 4.5?

**Comment by stepan-a** [comment]
I would like to have it in 4.5 (this would be of great help for nonlinear filters). We would need to agree on an interface and add an entry and references in the ref. manual.

**Comment by JohannesPfeifer** [comment]
@stepan-a That would be fine with me. I sent Marco a patch for including TaRB. This raises on issue, though. The preprocessor should directly set ```options_.posterior_sampling_method='tailored_random_block_metropolis_hastings'; ```
if the ```use_tarb``` option is specified. Because TaRB was only implemented in the unstable branch, we should probably replace the current ```use_tarb``` by a more expressive and flexible option like posterior_sampling_method.

**Comment by stepan-a** [comment]
@JohannesPfeifer I agree we can change the interface triggering TaRB. @rattoma and @JohannesPfeifer can you list the needed options here, so that @houtanb can implement the change in the preprocessor?

File: tests/estimation/TaRB/fs2000_tarb.mod [hunk 4]
```diff
@@ -114,7 +114,7 @@ end;
 
 varobs gp_obs gy_obs;
 
-estimation(order=1, datafile='../fsdat_simul',nobs=192, loglinear, mh_replic=30, mh_nblocks=2, mh_jscale=0.5,mode_compute=4,
+estimation(order=1, datafile='../fsdat_simul',nobs=192, loglinear, mh_replic=100, mh_nblocks=2, mh_jscale=0.5,mode_compute=4,
```
JohannesPfeifer: @rattoma I just realized that this change in mh_replic should not be included in the patch. Is it possible to amend this commit?

rattoma: Done with a forced push on my personal slice branch

-- 
Marco Ratto,
Financial and Economic Analysis
Joint Research Centre
European Commission,
TP 361, 21027  ISPRA(VA), ITALY
Tel: +39 0332 78 3794 Fax: +39 0332 78 5752,
marco.ratto@jrc.ec.europa.eu
http://www.macfinrobods.eu/

rattoma: done

matlab/check_posterior_sampler_options.m
```diff
@@ -76,17 +83,62 @@
...
               case 'rotated'
+                % triggers rotated slice iterations using a covariance
+                % matrix from initial burn-in iterations
+                % must be associated with:
+                % <use_mh_covariance_matrix> or <slice_initialize_with_mode>
+                % default  = 0
                 posterior_sampler_options.rotated = options_list{i,2};
+                
               case 'mode'
-                posterior_sampler_options.mode = options_list{i,2};
```
rattoma: Concerning this mode option, unless it is straightforward to implement it, I would simply cut it

houtanb:
pulled manually. keeping `slice` branch in `origin` so that we can work on #1177 there before merging into `master`

PR #1188, opened 2016-05-15
Add slice and factorization of posterior samplers

JohannesPfeifer:
The testsuite runs fine with the current state.
TBD:
- Dealing with ```adaptive_metropolis_hastings``` which now is broken as it relies on ```random_walk_metropolis_hastings_core.m``` which is supposed to be deleted soon and does not run anymore after the changes (@MichelJuillard  @stepan-a Do we have a mod-file triggering this part of the code somewhere)


JohannesPfeifer:
@stepan-a What should we do with ```adaptive_metropolis_hastings```? It is broken and seems to never have worked. None of the options used in the file exist and I could not find any documentation for what it actually does. I would propose removing it.

stepan-a:
@JohannesPfeifer I tend to agree with you, but this was developed by Michel... @MichelJuillard do you plan to finish/fix ```adaptative_metropolis_hastings```?

JohannesPfeifer:
@stepan-a Ok. I can discuss this with @MichelJuillard today over dinner. @rattoma and I think that this branch is now ready for merging.

PR #1196, opened 2016-05-20
OSR fixes related to parameter bounds

### LINKED ISSUE #948: Potentially add interface to set parameter bounds for osr.
In principle, there is everything in place to use constrained optimizers with osr. An example on how to use this with a user-defined optimizer wrapper is at http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=6835
However, we do not provide an interface to set such bounds. We should discuss if we should leave it at that (more work for users) or whether we provide functionality to automatically set the bounds (including their correct ordering in the parameter vector).

JohannesPfeifer:
Closes #948


JohannesPfeifer:
@stepan-a This PR also fixes the preprocessor error in OSR that makes the unit tests crash

stepan-a:
Manually merged.

PR #1214, opened 2016-06-05
Save filter_covariance and trends in Bayesian estimation

JohannesPfeifer:
- Saves trend and constant term in smoother after MCMC (Closes #1147)
- Save `oo_.Smoother.TrendCoeffs` for all variables instead of just observed (First part of #1162)
- Enables `filter_covariance` after MCMC (Closes #1160)

PR #1216, opened 2016-06-07
Harmonize FilteredVariables with Updated and Smoothed ones, by adding…

rattoma:
… steady state to stored values, both point estimate and MCMC


JohannesPfeifer:
@rattoma I don't think your are handling the `prefilter` part correctly in `prior_posterior_statistics_core`. We should probably wait for the merging of #1214, which stores the constant term explicitly after MCMC and makes adding it to filtered variables straightforward.

rattoma:
OK thanks. Let's take note that `filt_mc_` will have to be adapted accordingly as well. I will re-set my master branch for the moment.

rattoma:
@JohannesPfeifer I updated my PR.

stepan-a:
Manually merged.

PR #1217, opened 2016-06-07
Make sure diffuse_filter triggers qz_criterium>1

### LINKED ISSUE #1215: qz_criterium in estimation
I think it would be useful to allow option `qz_criterium` also in estimation, similarly to simulation. Would it be possible? Is there any strong reason for not allowing this?

JohannesPfeifer:
Closes #1215

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -6954,6 +6954,12 @@ more details.
 @item filter_decomposition
 @xref{filter_decomposition}.
 
+@item diffuse_filter = @var{INTEGER}
+@xref{diffuse_filter}.
+
+@item diffuse_kalman_tol = @var{DOUBLE}
+@xref{diffuse_kalman_tol}.
+
 @end table
 
 @end deffn
--- a/matlab/dynare_estimation_1.m
+++ b/matlab/dynare_estimation_1.m
@@ -31,6 +31,9 @@ function dynare_estimation_1(var_list_,dname)
 
 global M_ options_ oo_ estim_params_ bayestopt_ dataset_ dataset_info
 
+%store qz_criterium
+qz_criterium_old=options_.qz_criterium;
+
 % Set particle filter flag.
 if options_.order > 1
     if options_.particle.status && options_.order==2  
@@ -163,6 +166,8 @@ if isequal(options_.mode_compute,0) && isempty(options_.mode_file) && options_.m
         [atT,innov,measurement_error,updated_variables,ys,trend_coeff,aK,T,R,P,PK,decomp,Trend] = DsgeSmoother(xparam1,gend,transpose(data),data_index,missing_value);
         [oo_]=store_smoother_results(M_,oo_,options_,bayestopt_,dataset_,dataset_info,atT,innov,measurement_error,updated_variables,ys,trend_coeff,aK,P,PK,decomp,Trend);
     end
+    %reset qz_criterium
+    options_.qz_criterium=qz_criterium_old;
     return
 end
 
@@ -421,6 +426,8 @@ if (any(bayestopt_.pshape  >0 ) && options_.mh_replic) || ...
     %% Here I discard first mh_drop percent of the draws:
     CutSample(M_, options_, estim_params_);
     if options_.mh_posterior_mode_estimation
+        %reset qz_criterium
+        options_.qz_criterium=qz_criterium_old;
         return
     else
         if ~options_.nodiagnostic && options_.mh_replic>0
@@ -467,6 +474,8 @@ if (any(bayestopt_.pshape  >0 ) && options_.mh_replic) || ...
 end
 
 if options_.particle.status
+    %reset qz_criterium
+    options_.qz_criterium=qz_criterium_old;
     return
 end
 
@@ -713,3 +722,5 @@ if np > 0
     save([M_.fname '_pindx.mat'] ,'pindx');
 end
 
+%reset qz_criterium
+options_.qz_criterium=qz_criterium_old;
--- a/matlab/dynare_estimation_init.m
+++ b/matlab/dynare_estimation_init.m
@@ -123,29 +123,7 @@ if isequal(options_.diffuse_filter,1) || (options_.kalman_algo>2)
     end
 end
 
-% If options_.lik_init == 1
-%     set by default options_.qz_criterium to 1-1e-6
-%     and check options_.qz_criterium < 1-eps if options_.lik_init == 1
-% Else
-%     set by default options_.qz_criterium to 1+1e-6
-if isequal(options_.lik_init,1)
-    if isempty(options_.qz_criterium)
-        options_.qz_criterium = 1-1e-6;
-    elseif options_.qz_criterium > 1-eps
-        error(['Estimation: option qz_criterium is too large for estimating ' ...
-               'a stationary model. If your model contains unit roots, use ' ...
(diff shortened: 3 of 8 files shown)


stepan-a:
With this commit ```kalman/likelihood_from_dynare/fs2000_uncorr_ME.mod``` and ```kalman/likelihood_from_dynare/fs2000_uncorr_ME_missing.mod``` fail. I have the impression that the problem is more general than what is added in this commit when we run multiple ```estimation``` commands in the same mod file. The log of the first test is available [here](https://gitlab.com/snippets/21164).

JohannesPfeifer:
@stepan-a How about the solution in the commit I just attached?

stepan-a:
@JohannesPfeifer It's worse with your last patch. Now we have twenty failing integration tests:
``` 
     - conditional_forecasts/2/fs2000_est.mod
     - kalman/likelihood_from_dynare/fs2000_corr_ME.mod
     - kalman/likelihood_from_dynare/fs2000_corr_ME_missing.mod
     - lmmcp/sw_newton.mod
     - observation_trends_and_prefiltering/MCMC/Trend_loglin_no_prefilt_first_obs_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_loglin_prefilt_first_obs_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_loglinear_no_prefilter_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_loglinear_prefilter_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_no_prefilter_first_obs_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_no_prefilter_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_prefilter_first_obs_MC.mod
     - observation_trends_and_prefiltering/MCMC/Trend_prefilter_MC.mod
     - observation_trends_and_prefiltering/ML/Trend_loglinear_no_prefilter.mod
     - observation_trends_and_prefiltering/ML/Trend_loglinear_no_prefilter_first_obs.mod
     - observation_trends_and_prefiltering/ML/Trend_loglinear_prefilter.mod
     - observation_trends_and_prefiltering/ML/Trend_loglinear_prefilter_first_obs.mod
     - observation_trends_and_prefiltering/ML/Trend_no_prefilter.mod
     - observation_trends_and_prefiltering/ML/Trend_no_prefilter_first_obs.mod
     - observation_trends_and_prefiltering/ML/Trend_prefilter.mod
     - observation_trends_and_prefiltering/ML/Trend_prefilter_first_obs.mod

```
The first one, for instance, fails because when you reset ```options_``` in ```dynare_estimation``` we loose the value of ```qz_crtiterium``` which becomes empty.

JohannesPfeifer:
@stepan-a But this is a more general problem with us initializing `qz_criterium` as empty, not with the patch. I think it is a bad design choice for functions like `imcforecast` to rely on a value previously set in `dynare_estimation` or `stoch_simul`, but not be able to function with the default value.

stepan-a:
@JohannesPfeifer I agree that the design for ```qz_criterium``` is bad and unduly complex. The patch is faulty only w.r.t this peculiar design... From what I saw the problem is not only related to this option. The files under ```observation_trends_and_prefiltering``` seem to fail for another reason (another field in ```options_```?) I did not investigate.

Ideally the options should always be specific to the commands (_ie_ with different fields in ```options_```).

JohannesPfeifer:
@stepan-a There are two problems here:
1. How do we make sure `qz_criterium` is correctly reset? I would suggest to add a check for `isempty(options_.qz_criterium)` (as in e.g, `stoch_simul` and `osr`) to all entry-level routines where it is missing (like `imc_forecast`).
2. How to deal with `dynare_estimation_1` resetting options and thereby creating interactions with other commands? This is the reason e.g. the `observation_trends_and_prefiltering` fail, because `options_.nk` is reset based on `options_.filter_step_ahead`. Here, I would suggest to keep the current resetting of `options_` in `dynare_estimation_1` in place and then explicitly set all required options that need to be passed back at the end of `dynare_estimation_1`. That way it would be transparent what is reset within the routine.

JohannesPfeifer:
@stepan-a Any thoughts on this?

stepan-a:
@JohannesPfeifer Sorry I cannot work on this before next Thursday. I agree with the first point, we should write a routine to set/reset this option, and call this function on top of all the routines where this option is used (we need to identify them). Second point is less clear to me...

JohannesPfeifer:
Ok. 
The second point is about `dynare_estimation_1` resetting various global options. Some changes including e.g. `qz_criterium` are not supposed to be passed out after estimation because they create unintended interactions with other commands. Therefore, I introduced a resetting of the original options of the form
``` 
options_old=options_;
...
options_=options_old;
```
But this creates problems with options that are intended to be passed back like `nk`. What I am suggesting is to now do

``` 
options_old=options_;
...
options_old.nk=options_.nk;
options_=options_old;
```
That way we explicitly decide with options we want to pass back. This mirrors the approach taken in `stoch_simul`

JohannesPfeifer:
@stepan-a I reworked the PR according to our discussion and also added and tested the diffuse calibrated smoother. Please try to run the testsuite on this.

stepan-a:
There is still a problem with the last patch:
```bash
================================
DYNARE MAKE CHECK MATLAB RESULTS
================================
| TOTAL: 697
|  PASS: 685
|  FAIL: 6
| XFAIL: 6
| XPASS: 0
| LIST OF FAILED TESTS:
|     * conditional_forecasts/2/fs2000_est.mod
|     * kalman_filter_smoother/compare_results_simulation/fs2000_ML.mod
|     * shock_decomposition/fs2000_cal_groups.mod
|     * observation_trends_and_prefiltering/ML/Trend_loglinear_no_prefilter.mod
|     * observation_trends_and_prefiltering/ML/Trend_loglinear_no_prefilter_first_obs.mod
|     * observation_trends_and_prefiltering/ML/Trend_no_prefilter.mod
|
| LIST OF 10 SLOWEST TESTS:
|     * optimizers/fs2000_6.mod - 1439.220000
|     * observation_trends_and_prefiltering/MCMC/Trend_loglin_no_prefilt_first_obs_MC.mod - 1428.020000
|     * observation_trends_and_prefiltering/MCMC/Trend_loglinear_no_prefilter_MC.mod - 1377.670000
|     * observation_trends_and_prefiltering/MCMC/Trend_no_prefilter_MC.mod - 1341.400000
|     * observation_trends_and_prefiltering/MCMC/Trend_loglinear_prefilter_MC.mod - 1230.160000
|     * observation_trends_and_prefiltering/MCMC/Trend_prefilter_MC.mod - 1139.120000
|     * observation_trends_and_prefiltering/MCMC/Trend_prefilter_first_obs_MC.mod - 1114.580000
|     * observation_trends_and_prefiltering/MCMC/Trend_loglin_prefilt_first_obs_MC.mod - 1085.570000
|     * observation_trends_and_prefiltering/MCMC/Trend_no_prefilter_first_obs_MC.mod - 1064.900000
|     * ep/rbc_mc.mod - 1002.720000
```
For instance `observation_trends_and_prefiltering/ML/Trend_loglinear_no_prefilter.mod` fails in `imcforecast` because `qz_criterium` is empty.

JohannesPfeifer:
@stepan-a I adjusted the PR. It seems to work now on Sedna.

PR #1220, opened 2016-06-07
Restore compatibility of MLE with shock_decomposition and conditional_forecast

JohannesPfeifer:
Parameters are saved into `oo_.mle_mode` instead of `oo_.posterior_mode`, but functions did not account for this


stepan-a:
Manually merged.

PR #1221, opened 2016-06-10
Increase default MaxNumberOfBytes to 100 MB

### LINKED ISSUE #330: ReshapeMatFiles
ReshapeMatFiles is currently used only for posterior IRF's. For large models it is a huge bottle-neck which is probably useless. The approach followed in prior_posterior_statistics.m is to use the pm3 utility that takes the data needed for plots from the non-reshaped files. Such a procedure is in the end much less time consuming than the ReshapeMatFile.

The current pm3 utility is not easily extended for irfs (matrices are 4-dimensional for irf's).
However I tried to  write a pm4 utility for 4-D matrices which looks interesting.

How about removing the call in posteriorIRF to: 
ReshapeMatFile
posteriorIRF_core2 (the plot utility for bayesian irf's which uses the reshaped files)

add making a new call to a pm4.m utility that replicates the behaviour of pm3 to  4-D matrices?

JohannesPfeifer:
Closes #330


JohannesPfeifer:
Closed by e2942662479f822cd4ca1251158f0b5037c3a446

PR #1223, opened 2016-06-11
Transform automatic loading of optimal_mh_scale_parameter.mat into option scale_file

### LINKED ISSUE #503: Check use and precedence of jscale and document it.
The use of ``` bayestopt_.mh_jscale ``` and ``` bayestopt_.jscale ``` as well as ``` options_.mh_jscale ``` is confusing an potentially buggy (see also http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=5072). In ``` dynare_estimation_1 ``` we use 
``` 
mh_scale_fname = [M_.fname '_optimal_mh_scale_parameter.mat'];
if exist(mh_scale_fname)
    if options_.mode_compute == 0
        tmp = load(mh_scale_fname,'Scale');
        bayestopt_.mh_jscale = tmp.Scale;
        clear tmp;
    else
        % remove the file if mode_compute ~= 0
        delete('mh_scale_fname')
    end
end
``` 
to read out the ``` mh_jscale ``` from a previous run of ``` mode_compute=6 ```. But this is after ``` dynare_estimation_init ``` where the ``` mh_jscale ``` seems to be translated into the ``` bayestopt_.jscale ``` actually used in estimation. Thus, it looks as if this statement does nothing.

JohannesPfeifer:
Closes #503

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -5106,15 +5106,21 @@ The fraction of initially generated parameter vectors to be dropped as a burnin 
 
 @item mh_jscale = @var{DOUBLE}
 @anchor{mh_jscale}  The scale  parameter of  the jumping  distribution's
-covariance matrix (Metropolis-Hastings algorithm).  The default value is
-rarely satisfactory.  This  option must be tuned to  obtain, ideally, an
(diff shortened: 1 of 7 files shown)


stepan-a:
Manually merged.

PR #1235, opened 2016-06-16
Fix various Kalman filter bugs

### LINKED ISSUE #1231: Fix remaining Kalman-filter problems after #738
In particular, the univariate filter may return wrong results with measurement error
TBD from #738 
- Fix all bugs that result in unit tests crashing
- Improve unit tests to compare likelihood results for consistency
- Update and enable other unit tests for Kalman filter in kalman/likelihood directory

Please assign the ticket to me. Milestone should be 4.5

### LINKED ISSUE #1234: Check whether fast_kalman_filter is compatible with non-stationary states
From the Herbst (2015)-paper:
`This seems like a difficult problem, because the factorization is not obvious, and, in our
experience, even finding α consistently can be hard. Fortunately, the problem becomes
much easier once it is known that the states are stationary.`
In a nonstationary model I get a complex likelihood.

JohannesPfeifer:
- Fix bug in dsge_likelihood.m where diagonal of measurement error covariance matrix was not correctly read out, leading to wrong results when using the univariate Kalman filter with measurement error
- Various bugfixes for univariate smoother when (correlated) measurement error is present
- Bugfix for `missing_observations_kalman_filter_d.m` where constant of `log(2*pi)` was forgotten in last step
- Documents Kalman routines
- Factorizes and improves unit tests
- Filters out infinite likelihoods with proper error code
- Fixes variable count bug in `display_problematic_vars_Jacobian.m` that led to crashes in `debug`-mode

Closes #1231 

Closes #1234


JohannesPfeifer:
@stepan-a Could you please try `tests/kalman_filter_smoother/algo2`? It works on my machine but crashes on Sedna and I don't understand why.

MichelJuillard:
@JohannesPfeifer the 2*pi constant is a difficult issue in diffuse Kalman filter. If I remember correctly, because the diffuse prior is not a proper prior on initial values, the marginal density can't be compared between models and it is illusory to search for the "correct" constant. Sorry, this is a little fuzzy in my memory. Or maybe @JohannesPfeifer, do you only put the code in conformity with Durbin's formula?

JohannesPfeifer:
@MichelJuillard The formula is simply incorrectly implemented in the second branch of the if-clause. Formula 7.4 of DK (2012) clearly states that in any case the 2*pi must be added, but we do so in only one branch. This explains why even for the same model where no diffuse step are necessary the results differ

MichelJuillard:
OK then

rattoma:
@JohannesPfeifer thanks a lot for spotting this

JohannesPfeifer:
@stepan-a Never mind. My mistake, the testsuite works.

stepan-a:
Manually merged.

PR #1238, opened 2016-06-19
Bug fixes for identification with ML

### LINKED ISSUE #1237: Investigate identification problem
http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=8234
Sent email with mod-file to @rattoma on June 1, 2016

rattoma:
This fixes #1237


JohannesPfeifer:
@rattoma If I understand it correctly, these issue have been present in 4.4.3. Therefore, we need a bug report. Is the following a correct summary?

- When using ML, the asymptotic Hessian was not computed
- Checking for singular values when the eigenvectors contained only one column did not work correctly

rattoma:
this is perfect thank you

PR #1239, opened 2016-06-19
Make forecasts account for Measurement error

### LINKED ISSUE #853: Deal with forecast uncertainty due to measurement error
Currently, Bayesian forecasts completely neglect the presence of measurement error. Theoretically, is is not clear whether we should display
1. the true value of the underlying variable
2. the observed value of the variable including measurement error
While the mean forecast should be unaffected, the presence of measurement error should widen the uncertainty associated with forecasts. But this is in no way reflected in our forecasts. Additionally, neglecting the effect of measurement error on forecasts breaks the equivalence of specifying measurement error as a "structural" shock vs. specifying it using the Dynare command. 

At a minimum, we should document the treatment.

JohannesPfeifer:
- Closes #853 
- Introduce LaTeX Figure scaling by `\textwidth`
- Various smaller LaTeX fixes related to IRFs and GSA
- Improves x-axis ticks for IRFs


stepan-a:
I fixed two bugs (in ```forecast_graphs``` and ```forcst``` , I am waiting for the testsuite results for merging in master...

PR #1242, opened 2016-06-24
Expand functionality of rplot

JohannesPfeifer:
- Allow rplot to plot exogenous variables
 - Add eps-loader
 - Fix bug in figure saving


stepan-a:
Manually merged.

PR #1248, opened 2016-07-03
Fixes related to stack_solve_algo=7

JohannesPfeifer:
- Prevent using str2func on function handle
- Do not display debugging info for `perfect_foresight_solver` in `lnsrch1.m`
- Make `csolve` compatible with used sparse matrices
- Adjust unit test
- Use correct termination criterion in `dynare_solve'
- Provide interface to original PATH files


stepan-a:
Manually merged.

PR #1258, opened 2016-07-22
Restore backward compatibility of OSR command with maxit and tolf opt…

JohannesPfeifer:
…ions

Translates these options into optim_opt structure

### DIFF
--- a/matlab/osr1.m
+++ b/matlab/osr1.m
@@ -62,6 +62,27 @@ if any(isnan(M_.params(i_params)))
     error ('OSR: At least one of the initial parameter values for osr_params is NaN') ;
 end
 
+%restore backward compatibility with maxit and tolf
+if isfield(options_.osr,'maxit') || isfield(options_.osr,'tolf')
+    warning('OSR: The use of maxit and tolf is now deprecated. Please use the optim-option instead.')
+    if options_.osr.opt_algo~=4
+        error ('OSR: The maxit and tolf options are not supported when not using the default opt_algo=4. Use the optim-option instead.') ;
+    else
+        if isfield(options_.osr,'maxit')
+            if ~isempty(options_.optim_opt)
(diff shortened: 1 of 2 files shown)


matlab/osr1.m
```diff
@@ -62,6 +62,27 @@
     error ('OSR: At least one of the initial parameter values for osr_params is NaN') ;
 end
 
+%restore backward compatibility with maxit and tolf
+if isfield(options_.osr,'maxit') || isfield(options_.osr,'tolf')
+    fprintf('\nOSR: The use of maxit and tolf is now deprecated. Please use the optim-option instead.\n')
```
houtanb: Can this be changed to a `warning`?

houtanb:
@JohannesPfeifer I had understood that these options would no longer be included in the manual. Otherwise, for consistency, we would need to keep forever all deprecated options/commands in the manual.

JohannesPfeifer:
@houtanb Ok, I will remove the manual entry. But searching for `deprecated` in the manual looks as if we keep all options around.

JohannesPfeifer:
@houtanb I have done both changes

MichelJuillard:
The usual procedure is first to declare an option or an instruction as deprecated in the manual and indicate its replacement. But the option or instruction is still working.
Then, a few versions later, we remove the option or the instruction altogether. Version 5.0 will be a good occasion to remove features declared obsolete a long time ago and introduce explicit error messages.
The compatibility code that we are maintaining should support old features forever.

houtanb:
@MichelJuillard I totally agree with the process. My concern is that we never get to step two wherein we actually remove the entry, leaving things forever deprecated.

houtanb:
@JohannesPfeifer sorry for the extra work, but please add back the manual entry as per @MichelJuillard 's comment

JohannesPfeifer:
Ok. Done.

PR #1267, opened 2016-08-17
Fix two bugs in dynare_estimation_init.m

JohannesPfeifer:
Makes sure that i) name field is always present and ii) that number fields of estim_params are only accessed after they have been set by set_prior

### DIFF
--- a/matlab/dynare_estimation_init.m
+++ b/matlab/dynare_estimation_init.m
@@ -142,7 +142,7 @@ else
 end
 
 % Set priors over the estimated parameters.
-if ~isempty(estim_params_) && ~(isfield(estim_params_,'nvx') && sum(estim_params_.nvx+estim_params_.nvn+estim_params_.ncx+estim_params_.ncn+estim_params_.np)==0)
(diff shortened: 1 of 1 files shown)


MichelJuillard:
If this correction was necessary, it seems that something else is broken upstream in the preprocessor or the functions calling dynare_estimation_init

JohannesPfeifer:
@MichelJuillard No, this was my mistake. The particular problem here was that I erroneously accessed e.g. `estim_params_.nvx`, which is only constructed within the if-clause from `size(estim_params_.var_exo,1)`. Thereby, it was possible to access the  stale `estim_params_.nvx` from previous calls to `calib_smoother`.

MichelJuillard:
OK thanks

MichelJuillard:
@JohannesPfeifer this commit breaks some test cases: you can't count on the existence of any estim_params_ fields except nvx nvn ncx ncn and np

JohannesPfeifer:
@MichelJuillard I see. I pushed a commit to #1270.

PR #1268, opened 2016-08-19
Remove error-display from dynare_solve.m

JohannesPfeifer:
Prevents estimation crashing when invalid starting values for steady state computation arise during e.g. mode-finding

### DIFF
--- a/matlab/display_problematic_vars_Jacobian.m
+++ b/matlab/display_problematic_vars_Jacobian.m
@@ -33,6 +33,7 @@ function []=display_problematic_vars_Jacobian(problemrow,problemcol,M_,x,type,ca
 % You should have received a copy of the GNU General Public License
 % along with Dynare.  If not, see <http://www.gnu.org/licenses/>.
 
+skipline();
 if nargin<6
     caller_string='';
 end
@@ -114,8 +115,8 @@ elseif strcmp(type,'static')
         end
     end
     fprintf('\n%s  The problem most often occurs, because a variable with\n',caller_string)
-    fprintf('%s exponent smaller than 1 has been initialized to 0. Taking the derivative\n',caller_string)
-    fprintf('%s and evaluating it at the steady state then results in a division by 0.\n',caller_string)
+    fprintf('%s  exponent smaller than 1 has been initialized to 0. Taking the derivative\n',caller_string)
+    fprintf('%s  and evaluating it at the steady state then results in a division by 0.\n',caller_string)
     fprintf('%s  If you are using model-local variables (# operator), check their values as well.\n',caller_string)
 else
     error('Unknown Type')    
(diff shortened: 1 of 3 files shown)


MichelJuillard:
When the problem occurs, you return info=1, this is not the proper code in print_info

JohannesPfeifer:
@MichelJuillard As far as I can see, the error handling is not done in `dynare_solve`, but later functions like `evaluate_steady_state`. `dynare_solve` always only returns 0 or 1.

MichelJuillard:
Yes, of course, dynare_solve is now a generic function for general nonlinear system of equation, not only for computing the steady state. But then, in the debug message, we should remove references to steady or static

JohannesPfeifer:
@MichelJuillard Then we should probably get rid of the debug message, because there is no way to ascertain on which function dynare_solve is called and the call to `display_problematic_vars_Jacobian` will return wrong results.

MichelJuillard:
OK, let's get rid of it.

JohannesPfeifer:
@MichelJuillard I amended the commit to only provide the debugging output for steady state computation. That should solve the issue.

MichelJuillard:
I don't see the change in the commit

JohannesPfeifer:
@MichelJuillard The new part is the test whether the caller is evaluate_steady_state in line 56 of 444cf1735e98183adf91da03c90eb67ac199025f

JohannesPfeifer:
@MichelJuillard I did another round in 	94128bd. I would like to keep the call to `display_problematic_vars_Jacobian` for the steady state case. because I rely heavily on it with mod-files of users in the forum.

JohannesPfeifer:
@MichelJuillard Ok. In 5e4f52e542c3a738d8d2e4ea2f70f456edb15556 I removed all application-specific output.

MichelJuillard:
OK for dynare_solve. Please, split the PRs.  In evaluate_steady_state:
1) we don't want to bring back fvec and fjac in normal circumstances, only if there is an error flag. Repeat the call to dynare_solve requesting fvec and fjac inside the debug block
2) You call the debug block even if there is no error in the return of dynare_solve

JohannesPfeifer:
@MichelJuillard I would prefer to always bring back `fvec` and `fjac`. As they are always computed by `dynare_solve` regardless of whether we output them, this should be a costless operation. In contrast, calling `dynare_solve` a second time to recompute the Jacobian is really expensive when someone runs estimation in debug mode.

MichelJuillard:
@JohannesPfeifer we don't care about expensive when debug is on and an error has been caught
Returning a matrix is not costless in Matlab, because space is allocated and the matrix is copied. For large Jacobian, it is a useless operation in normal circumstances

JohannesPfeifer:
@MichelJuillard Ok. Done.

PR #1275, opened 2016-08-31
Replace hline in dyn_latex_table.m by proper booktabs style

JohannesPfeifer:


matlab/rplot.m
```diff
@@ -101,6 +102,7 @@ function rplot(s1)
     for j = 1:size(y,1)
         hh=dyn_figure(options_,'Name',['Simulated Trajectory']);
         plot(ix(i),y(j,i)) ;
+        xlim([min(ix(i)) max(ix(i))])
```
houtanb: Does `xlim` not apply when `rplottype == 2`?

JohannesPfeifer:
@houtanb No, because we have an `axis tight` command in that case.

PR #1279, opened 2016-09-05
make computations around invhess more robust numerically.

rattoma:
for example removes many occurrences of `-Inf` in computing Laplace approximations by scaling invhess for computing log of determinant.

### DIFF
--- a/matlab/dynare_estimation_1.m
+++ b/matlab/dynare_estimation_1.m
@@ -339,7 +339,8 @@ if ~options_.mh_posterior_mode_estimation
         oo_.posterior.optimization.log_density=-fval;
     end
     if options_.cova_compute
-        invhess = inv(hh);
+        hsd = sqrt(diag(hh));
+        invhess = inv(hh./(hsd*hsd'))./(hsd*hsd');
         stdh = sqrt(diag(invhess));
         oo_.posterior.optimization.Variance = invhess;
     end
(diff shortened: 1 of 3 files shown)


stepan-a:
@rattoma ```estimation/fs2000_model_comparison``` fails on this branch. The inversion formula affects too much the evaluation of the Laplace approximation. The difference with the harmonic mean estimator is greater (by far) than the admitted tolerance (0.2)... I do not know of this is a problem. Maybe we should use a model for which we have a closed form expression for the posterior marginal density (or its Laplace approximation) as a test.

rattoma:
@stepan-a thanks a lot.  I will look into this

rattoma:
@stepan-a, invhessian and Laplace are invariant to my transformation. My transformation just makes computations more robust for large models, where master may give Inf for log data density. For standard models, results are identical up to the 13th significant digit.

Concerning your findings, the issue is that the test is wrongly designed.

In the master branch, `estimation/fs2000_model_comparison` does not fail because Laplace approximation is `NaN`, due to the last call to estimation in `fs2000.mod`.
More precisely, in that call `hh` is replaced by an identity matrix:
```
hh=eye(size(bayestopt_.name,1));
save('fs2000_mode','hh','-append')
```
which, in master branch creates in `dynare_estimation_1` line 368,
```
        scale_factor = -sum(log10(diag(invhess)));
``
`scale_factor=0`, which goes in next line producing `log(0)`.

In my branch I get rid of that scale factor, and the computation of Laplace approximation is overwritten not by NaN, but by a real value that uses identity as invhess in place of the actual invhess. This makes Laplace strongly different w.r.t. ModifiedHarmonicMean.

After the first call to estimation in `fs2000.mod`, I get the following in my branch:
```
    LaplaceApproximation: 1.186686997856009e+03
    ModifiedHarmonicMean: 1.186277614656287e+03
```
which is the same as in master:
```
    LaplaceApproximation: 1.186686997856009e+03
    ModifiedHarmonicMean: 1.186277614656288e+03
```

I am going to add a modification to `fs2000.mod` in PR that fixes this. Keeping the 0.2 threshold for the difference, BOTH branches now fail. I also raised the threshold to 0.5 for the diff between Laplace and MHM estimatio, which makes the test OK.

rattoma:
here is the addition to my PR for the testsuite
e46a3249c01189c1d345011814551eb22eec4555

stepan-a:
Thanks for the detailed explanation @rattoma.

PR #1281, opened 2016-09-12
Fixed bug with new harmonized setting of Smoothed Variables, which co…

rattoma:
…ntain already steady-state also at the mode without metropolis


JohannesPfeifer:
@houtanb Could you please also pull this?

PR #1295, opened 2016-10-02
Stochastic singularity

### LINKED ISSUE #1294: Introduce proper check for stochastic singularity
In `initial_estimation_checks` we should introduce a proper check for stochastic singularity by setting `options_.use_univariate_filters_if_singularity_is_detected=0` and then providing an informative error message when `info(1) == 50`.
This prevents user errors where the model has a fundamental stochastic singularity, because Dynare by default will resort to the univariate filter without warning.

JohannesPfeifer:
Closes #1294

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -6020,6 +6020,15 @@ the option is equal to @code{0}, Dynare will not automatically change
 the filter, but rather use a penalty value for the likelihood when
 such a singularity is encountered. Default: @code{1}.
 
+@item keep_kalman_algo_if_singularity_is_detected
+@anchor{keep_kalman_algo_if_singularity_is_detected}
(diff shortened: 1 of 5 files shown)


JohannesPfeifer:
@MichelJuillard I personally have never encountered a case like this (unless a shock variance is set to 0 by mistake). But as the check is for the initial parameters and the data is a given, the logical step is to use different initial parameters. Do you think we should indicate this? I abstained from doing so, because I think this is a rare problem.

MichelJuillard:
@JohannesPfeifer If we mention the rare possibility, we should make a suggestion on how to handle it and add to the error message something like:
"If you think it is the later case, you may try with different initial values for the estimated parameters"

PR #1297, opened 2016-10-03
Bugfixes and cosmetic changes related to estimation

JohannesPfeifer:
- Solves Octave unit test fails related to `mh_recover` by conditioning the seed resetting on the presence of Octave
- Fixes bug in `GetPosteriorMeanVariance.m`when number of dropped draws coinided with number of draws in the mh_file; in this case the results were NaN

### DIFF
--- a/matlab/posterior_sampler_core.m
+++ b/matlab/posterior_sampler_core.m
@@ -126,9 +126,13 @@ for curr_block = fblck:nblck,
         % Matlab/Octave cluster). Therefore the trap.
         %
         % Set the random number generator type (the seed is useless but needed by the function)
-        set_dynare_seed(options_.DynareRandomStreams.algo, options_.DynareRandomStreams.seed);
+        if ~isoctave
+            set_dynare_seed(options_.DynareRandomStreams.algo, options_.DynareRandomStreams.seed);
+        else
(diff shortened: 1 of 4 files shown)


matlab/posterior_sampler_core.m
```diff
@@ -126,9 +126,15 @@
         % Matlab/Octave cluster). Therefore the trap.
         %
         % Set the random number generator type (the seed is useless but needed by the function)
-        set_dynare_seed(options_.DynareRandomStreams.algo, options_.DynareRandomStreams.seed);
-        % Set the state of the RNG
-        set_dynare_random_generator_state(record.InitialSeeds(curr_block).Unifor, record.InitialSeeds(curr_block).Normal);
+        if ~isoctave
+            set_dynare_seed(options_.DynareRandomStreams.algo, options_.DynareRandomStreams.seed);
+            % Set the state of the RNG
+            set_dynare_random_generator_state(record.InitialSeeds(curr_block).Unifor, record.InitialSeeds(curr_block).Normal);
```
houtanb: It seems like 132 is the same as 136. Can you move 131-132 outside of the if statement and delete 135-136

JohannesPfeifer:
@houtanb Thanks, I updated the commit

PR #1298, opened 2016-10-04
Transform save_tmp_file into an option and change defaults

### LINKED ISSUE #1296: Decide on whether to save intermediate draws
With the move to `posterior_sampling_core` we now by default save the MCMC draws every 50 draws into a temporary file, because we by default set
`posterior_sampler_options.save_tmp_file=1;`

I think this is very inefficient and therefore should be 0 by default (same behavior as before the move), with an interface provided to change the option.
@rattoma You added this behavior. Was the reason that `slice` should be treated differently?

JohannesPfeifer:
Closes #1296


stepan-a:
Manually merged.

PR #1301, opened 2016-10-05
Use raw_data in forecast plots instead of prefiltered data

### LINKED ISSUE #1299: The plot of recursive forecasts is wrong when prefilter=1
GNU Octave, version 3.8.1 "i686-w64-mingw32".
Dynare unstable (version 2016-08-23).

When the variables are prefiltered, the plot of recursive forecast is made mixing a filtered realized variable with a defiltered forecasted variable.
In the window below, this is very visible in the juros variable, but happens in all of them.

What I woud expect is all of them to be in the defiltered mode.

JohannesPfeifer:
Closes #1299

### DIFF
--- a/matlab/dynare_estimation.m
+++ b/matlab/dynare_estimation.m
@@ -29,7 +29,7 @@ function oo_recursive_=dynare_estimation(var_list,dname)
 % You should have received a copy of the GNU General Public License
 % along with Dynare.  If not, see <http://www.gnu.org/licenses/>.
 
-global options_ oo_ M_ dataset_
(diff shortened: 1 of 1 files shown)


JohannesPfeifer:
@stepan-a Is it intended that ony `dataset_info` stores the raw data. I would have preferred to have subfields `data` and `rawdata` in `dataset_`

stepan-a:
@JohannesPfeifer I do not remember. I guess I wanted to isolate what is strictly necessary for estimation in `dataset_`... I need to think about this.

PR #1323, opened 2016-10-27
Add unit test for qz_criterium

JohannesPfeifer:


stepan-a:
I reverted the merge in a60f5a81fed275dbfe8c7409ded0eda78e2d64a1. I do not understand the test which fails. @houtanb we should avoid to add new tests if they fail, they only add noise to the testsuite. Please do not merge without running the tests.

JohannesPfeifer:
@stepan-a Do you have a log somewhere? I tried the test and it worked for me.

stepan-a:
The test fails on the test at the end (```error_indicator``` is equal to 1). I wonder if this is related to the the version of matlab (I used 2015b when running the test, while the testsuite on sedna use the lastest version of matlab)... We would need to debug the test with both versions of matlab to identify the culprit. I won't do it now...

houtanb:
I am surprised this works for @JohannesPfeifer as `error_indicator` does not always exist. It only exists if we do not go into the catch portion of the block.

houtanb:
@stepan-a the test works fine. The problem is that you somehow modified your local copy (the one you posted in your `gist`) by removing the `error_indicator = 0;` line. I'm going to revert your revert as the test works correctly as originally committed

PR #1324, opened 2016-10-28
Three bugfixes for missing_DiffuseKalmanSmootherH1_Z.m that led to wrong results

JohannesPfeifer:
1. In case of missing observations ('if isempty(di)'), `a(:,t)` was not propagated forward to update `a(:,t+1)`
2. In the rank-deficient `Finf` case, `Kstar` was defined as `T^(-1)*K^{(0)}`, while in the full rank it was defined as `Kstar=K^(0)` This leads to wrong results when switches between the two clauses occurr. Moreover, the later backwards pass relied on `Kstar=K^(0)`, leading to wrong results when the rank-deficient `Finf` case was triggered earlier. 
3. In the rank-deficient `Finf` case, the backward pass did not use the proper formulas, but rather the ones for the non-singular case

The implementation now consistently follows the one in `kalman_filter_d.m`


matlab/missing_DiffuseKalmanSmootherH1_Z.m
```diff
@@ -83,71 +85,74 @@
...
                         alphahat = Inf;
                         return
-                    else
+                    else %rank 0
                         a(:,t+1) = T*a(:,t);
                         Pstar(:,:,t+1) = T*Pstar(:,:,t)*transpose(T)+QQ;
                         Pinf(:,:,t+1)  = T*Pinf(:,:,t)*transpose(T);
                     end
                 else
                     iFstar = inv(Fstar(:,:,t));
```
JohannesPfeifer: There seems to be a third bug here. `iFstar` only stores the matrix for every `t`, but the next line uses it with a `t` index. I will fix this when I am back at my computer.

PR #1340, opened 2016-11-19
Fix bug in missing_DiffuseKalmanSmootherH3_Z.m

### LINKED ISSUE #1339: bug in missing_DiffuseKalmanSmootherH3_Z.m?
@JohannesPfeifer I don't see how line 302 
```
                    Linf    = eye(mm) - Kinf(:,i,t)'/Finf(i,t);
```
can be correct. The term before - is a square matrix and the term after, a row vector

JohannesPfeifer:
Closes #1339 by properly factorizing computations


MichelJuillard:
@JohannesPfeifer I may wait to merge it until we know why the testsuite didn't catch it

PR #1346, opened 2016-12-01
Correct typos in error message and header

JohannesPfeifer:


matlab/dr_block.m
```diff
@@ -638,7 +638,7 @@
 
             
             if options_.loglinear
-                error('log linear option is for the moment not supported in first order approximation for a block decomposed mode');
+                error('loglinear option is for the moment not supported in first order approximation for a block decomposed model');
```
houtanb: Should read, "The loglinear option is not supported at this time in first order approximation for a block decomposed model"

JohannesPfeifer:
@houtanb Thanks for pointing this out. I now took the time to do a more encompassing change in the amended commit.

PR #1347, opened 2016-12-09
Bugfix for OSR

JohannesPfeifer:
- Fixes bug in `osr_obj`
- Suppresses OSR output if noprint option is specified
- Adds error indicator to get status when looping over OSR


stepan-a:
Manually merged.

PR #1348, opened 2016-12-09
model_diagnostics.m: provide warning that check of dynamic model is n…

JohannesPfeifer:
…ot supported for block option instead of crashing


stepan-a:
Manually merged.

Issue #1201, opened 2016-05-26
Depth issue


stepan-a:
Looking into #1175 , by reverting commit 3c7e60b744567f6f39a9c611bce6dcaadcd52bc6, I obtained the following error from matlab when trying to run Christiano-Motto-Rostagno model (the one in subfolder `figure4` of the archive available [here](http://faculty.wcas.northwestern.edu/~lchrist/research/ECB/risk_shocks/20100922_data.zip)

```
Error: File: cmr_static.m Line: 1292 Column: 16331
Nesting of {, [, and ( cannot exceed a depth of 32.
```

@MichelJuillard May this be a consequence of your patch about on auxiliary variables in steady state and static files (see #1133)?

MichelJuillard:
@stepan-a. This remind me of something. I take a look

MichelJuillard:
I believe now that this issue is related to another change that occurred when we separated the temporary terms between the computation of the residuals and the computation of derivatives as various order. I don't remember who did it and how the issue of the commit is called.

JohannesPfeifer:
Do you mean https://github.com/DynareTeam/dynare/commit/127637ffd672fdd8143ee13f2a38234e597d94d3

MichelJuillard:
Thanks @JohannesPfeifer. That is the one. I think @houtanb did it before @stepan-a merged it in master. I will contact @houtanb to see how to handle it

MichelJuillard:
Here is a short example. Let's call it mod1.mod:
```
var x y;
varexo e;

model;
y = (1/x(+1))*(1/x(+2))*(1/x(+3));
x = exp(e);
end;

steady_state_model;
y = 1;
x = 1;
end;

steady;

stoch_simul(order=1,irf=0);
```
Look at the temporary term at the start of Jacobian computation in mod1_static.m. If this expression used temporary terms for common terms, it would solve the depth problem when there are lots of terms in the denominator.

houtanb:
NB: This bug does not exist in 127637ffd672fdd8143ee13f2a38234e597d94d3. It has been introduced since then, and seems to belong to the `static` model, as the `dynamic` model is the same between `master` and 127637ffd672fdd8143ee13f2a38234e597d94d3. That said, the CMR model referenced above breaks for a reference to `logncdf` which is not found.

houtanb:
It seemed as though the bug came from `substituteStaticAuxiliaryDefinition()` but I could not immediately find the fault in the logic.

[...]

PR #1354, opened 2016-12-18
Use master function lyapunov_solver.m to call individual solvers

### LINKED ISSUE #821: Add flag if lyapunov_symm fails
In models like the one at http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=6291
`lyapunov_symm` fails to solve the Lyapunov equation and outputs NaN. This error is not flagged and no exception thrown out, leading to subsequent crashes. We should add a check whether the solver was successful and only continue if so.

### LINKED ISSUE #1077: Integrate doubling algorithm into lyapunov_symm
Following #988 and the discussions at the 2015 Dynare conference, we should integrate `disclyap_fast.m` into `lyapunov_symm` to have one solver only. The question is whether we want to integrate this as a call to `disclyap_fast.m`as an external function or whether to make `disclyap_fast.m` a nested function within  `lyapunov_symm`. I would prefer the latter, because it makes the function with the doubling algorithm reusable as a standalone in the way Herbst (2015) and Schmitt-Grohé/Uribe already use it.

When doing this, we also need to discuss the option handling as we currently do not pass the doubling tolerance to `lyapunov_symm`. Currently, we pass `lyapunov_fixed_point_tol`. I would propose to use this input argument as the tolerance criterion if the doubling algorithm is selected.

JohannesPfeifer:
Closes #821 by switching to Schur-based method by default when iteration fails. 
Closes #1077 by moving call to Matlab's `dlyapchol`  and to `disclyap_fast` to master function `lyapunov_solver`

### DIFF
--- a/matlab/DsgeSmoother.m
+++ b/matlab/DsgeSmoother.m
@@ -149,15 +149,7 @@ if options_.lik_init == 1               % Kalman filter
     if kalman_algo ~= 2
         kalman_algo = 1;
     end
-	if options_.lyapunov_fp == 1
-        Pstar = lyapunov_symm(T,R*Q*R',options_.lyapunov_fixed_point_tol,options_.qz_criterium,options_.lyapunov_complex_threshold, 3, [], options_.debug);
-    elseif options_.lyapunov_db == 1
-        Pstar = disclyap_fast(T,R*Q*R',options_.lyapunov_doubling_tol);
-    elseif options_.lyapunov_srs == 1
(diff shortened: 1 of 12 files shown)


JohannesPfeifer:
@stepan-a Do I see it correctly that you manually merged this ticket with c893c0620c620c9aefec349a7c8628595a9c0dbd?

stepan-a:
Yes. Sorry I forgot to close this thread... Playing with the nested parenthesis example (issue #1201)

PR #1357, opened 2016-12-19
Provisions for new node option NumberOfThreadsPerJob. Document option

### LINKED ISSUE #1349: set number of threads per job in parallel execution
Given the large number of CPU's being made available on desktop pc's,  the current node option `singleCompThread` looks rather limited.
I would propose the following changes to the parallel configuration:
- set the default for `singleCompThread` to 0;
- add a node option in dynare.ini `numberOfThreadsPerJob`, which allows to set the number of cpu's assigned to each job. So, for example, for a node with 8 CPU's and `numberOfThreadsPerJob=2`, the parallel engine will split the work into 4 parallel instances, each using 2 cpu's. 

would this be possible?

rattoma:
This commit closes #1349. thanks a lot

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -10815,6 +10815,10 @@ installation directory. The default is the empty string.
 The path to the MATLAB or Octave executable. The default value is
 @code{matlab}.
 
+@item NumberOfThreadsPerJob = @var{INTEGER}
+For Windows nodes, sets the number of threads assigned to each remote MATLAB/Octave run. The default
+value is @code{1}. 
+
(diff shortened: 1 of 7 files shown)


houtanb:
@rattoma is there a reason to keep the commented code in 4ecf1d9f5d3fdf6e9ddbe6f054bc5a8f8de9cd52?

rattoma:
@houtanb the commented lines would be an alternative solution, should users report problems with the saved figures. The current solution should work, however. Perhaps we should add a note to the commented lines?

rattoma:
I just added a note to commented lines in 4ecf1d9

PR #1358, opened 2016-12-20
Implement posterior_nograph option

### LINKED ISSUE #1356: introduce posterior_nograph option
I think it would be useful to have an option like:

`options_.posterior_nograph`

with default value at 0 like standard nograph. 

This option would allow performing all possible posterior subdraws for irfs, smoothed variables etc, but avoiding to produce thousands of useless graphs that are currently done [for large models at least]. 

I think it would also be useful to make a distinction w.r.t. `options_.nograph` since the latter would not trigger prior/posterior plots of parameter estimates which would be required in any case. 

And by the way, `pm3` and `posterior_irf` do not seem to honor the nograph option: plots are made in any case. So at least we should make sure `nograph` would work properly?

If you agree, I would make the appropriate changes to the matlab `pm3` and `posterior_irf` routines with `posterior_nograph` [the preprocessor could be updated with this extra option, then].

JohannesPfeifer:
Also finishes implementation and documentation of nograph option of shock_decomposition.m
Closes #1356

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -7030,6 +7030,11 @@ the decomposition. Groups of shocks are defined in @xref{shock_groups} block.
 @item colormap = @var{COLORMAP_NAME}
 @anchor{colormap}. Controls the colormap used for the shocks decomposition
 graphs. See @code{colormap} in Matlab/Octave manual.
+
+@item nograph
+@xref{nograph}. Suppresses the display and creation only within the @code{shock_decomposition}-command
+but does not affect other commands.
+
 @end table
 
 @vindex oo_.shock_decomposition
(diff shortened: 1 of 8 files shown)


JohannesPfeifer:
@rattoma I added the changes you sent me and the unit tests works.

rattoma:
@JohannesPfeifer many thanks!

houtanb:
@JohannesPfeifer You need to add the options `posterior_nograph` and `posterior_graph` to the `estimation` command in the manual.

Also, I would've split these commits up differently. 1 commit for the change in `nograph` for `shock_decomposition`. And another commit for the `posterior_graph` and `posterior_nograph` options, combining the preprocessor/manual changes with e0b6318f5483ec195ea8b5b57a28507b32b68127

JohannesPfeifer:
@houtanb Thanks. I adjusted the PR accordingly

Issue #1162, opened 2016-04-11
Handling of trends


MichelJuillard:
Currently trends are taken into account only if users indicate them for observed variables. However, in random walk with drift, a deterministic trend appears endogenously for variables that are not necessarily observed.

Let consider the following model where only Y is observed:
```
A_t = A_{t-1} + g + e_t
Y_t = A_t + eta_t
```
Both A_t and Y_t contain a deterministic linear trend with a slope of g. The current practice of only specifying the trend of Y is not satisfactory anymore.

When using the smoother, we need to recognize these trends and include them in SmoothedVariables and friends

JohannesPfeifer:
Solution agreed with @MichelJuillard: Provide subfields with size ```endo_vars``` for all variables and set trend to NaN for variables with unspecified trends. Allows for future setting of trends for unobserved variables using e.g. the ```trend_var``` command.

JohannesPfeifer:
As discussed in #1372, some dataset features may require information on trends/deflators. They currently access only `M_`. The question is then where to exactly store trend information required in various contexts

rattoma:
@JohannesPfeifer thank you. I think the point here may also go beyond dataset, but also to allow users to add trends to latent/unobserved variables that have been removed by `trend_var` in the model.

PR #1372, opened 2017-01-08
Shock decompositions: plotting utilities and realtime utility

rattoma:
I have completed the work on plot shock decompositions. I have added two new commands:
``
plot_shock_decomposition(OPTIONS) VAR_LIST;
``
that should be translated by the pre-processor as:
``
plot_shock_decomposition(M_,oo_,options_,var_list_);
``
and 
``
realtime_shock_decomposition(OPTIONS) VAR_LIST;
``
that should be translated by the pre-processor as:
``
oo_=realtime_shock_decomposition(M_,oo_,options_,var_list_,bayestopt_,estim_params_);
``
there are several new options. To minimize the workload on the preprocessor side, I think all these options could be set as NAME+VALUE instead of adding new pre-processor entries.
This would imply the following requirements for the preprocessor:
- trigger the two new commands `plot_shock_decomposition` and `realtime_shock_decomposition`. 
- Allow a new NAME+VALUE field in `options_`, like
``
options_.shock_decomp.decomp_opt
``
- I will then prepare the routine similar to `check_posterior_sampler_options.m` that treats these options.

If you think it is better to extend the pre-processor, I will send you the list of options.


doc/dynare.texi
```diff
@@ -366,13 +366,13 @@ are willing to bear the concomitant performance loss.
 The development of Dynare is mainly done at
 @uref{http://www.cepremap.fr, Cepremap} by a core team of
 researchers who devote part of their time to software development.
-Currently the development team of Dynare is composed of Stéphane
-Adjemian (Université du Maine, Gains and Cepremap), Houtan Bastani
-(Cepremap), Michel Juillard (Banque de France), Frédéric Karamé
-(Université du Maine, Gains and Cepremap), Junior Maih (Norges Bank),
-Ferhat Mihoubi (Université Paris-Est Créteil, Epee and Cepremap), George
+Currently the development team of Dynare is composed of St�phane
```
JohannesPfeifer: @rattoma For some reason your editor killed the special characters in the manual file

matlab/shock_decomposition.m
```diff
@@ -148,6 +148,6 @@
     shock_names = M_.exo_names;
 end
         
-if ~options_.no_graph.shock_decomposition
+if ~options_.nograph && ~options_.no_graph.shock_decomposition
```
JohannesPfeifer: @rattoma I am not sure this works as intended. The `shock_decomposition` command has a `nograph` option that sets `options_.no_graph.shock_decomposition`, but not `options_.nograph`. Should we really make `shock_decomposition` obey this option set only by other commands? If yes, we should probably go for a different name for the preprocessor option of the command itself, like `shock_decomposition_nograph`. Otherwise, it is confusing.

matlab/plot_shock_decomposition.m
```diff
@@ -0,0 +1,196 @@
+function plot_shock_decomposition(M_,oo_,options_,varlist)
+% function plot_shock_decomposition(M_,oo_,options_,varlist)
```
JohannesPfeifer: @rattoma It seems this header is copy and pasted from `shock_decomposition.m` and therefore wrong. As far as I can see, this function does not provide any output

matlab/realtime_shock_decomposition.m
```diff
@@ -0,0 +1,275 @@
...
+    M1=M_;
+    M1.params = forecast_params;
+    [junk1,junk2,junk3,junk4,junk5,junk6,oo1] = dynare_resolve(M1,options_,oo_);
+    clear junk1 junk2 junk3 junk4 junk5 junk6
+end
+
+for j=presample+1:nobs,
+%    evalin('base',['options_.nobs=' int2str(j) ';'])
+    options_.nobs=j;
+    clear('evaluate_smoother');
```
JohannesPfeifer: @rattoma Why do you need this `clear`?

matlab/realtime_shock_decomposition.m
```diff
@@ -0,0 +1,275 @@
+function oo_ = realtime_shock_decomposition(M_,oo_,options_,varlist,bayestopt_,estim_params_)
+% function z = realtime_shock_decomposition(M_,oo_,options_,varlist,bayestopt_,estim_params_)
```
JohannesPfeifer: @rattoma This header also seems to be copy and pasted. The output of this file seems to be different from what is stated here.

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
 
 @end deffn
 
+@deffn Command realtime_shock_decomposition [@var{VARIABLE_NAME}]@dots{};
+@deffnx Command realtime_shock_decomposition (@var{OPTIONS}@dots{}) [@var{VARIABLE_NAME}]@dots{};
+
+@descriptionhead
+
+This command computes the recursive historical shock decomposition for a given sample based on
```
JohannesPfeifer: @rattoma What is the difference between realtime and historical/normal shock decomposition. In this line you write that realtime computes the _historical_ decomposition, which does not sound like realtime.

doc/dynare.texi
```diff
@@ -7065,8 +7065,12 @@ graphs. See @code{colormap} in Matlab/Octave manual.
 
 @item nograph
 @xref{nograph}. Suppresses the display and creation only within the @code{shock_decomposition}-command
-but does not affect other commands.
+but does not affect other commands. See also @xref{plot_shock_decomposition} for plotting graphs.
 
+@item init_state = @var(INTEGER)
+@xref{init_state}. It can take values of 0 (=default) and 1.
```
JohannesPfeifer: What happens in case of 0? I thought we always start at the smoothed states?

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+calibrated model.
+
+@item first_obs = @var{INTEGER}
+@xref{first_obs}.
+
+@item nobs = @var{INTEGER}
+@xref{nobs}.
+
+@item use_shock_groups [= @var{SHOCK_GROUPS_NAME}]
+@anchor{use_shock_groups}. Uses groups of shocks instead of individual shocks in
```
JohannesPfeifer: @rattoma It seems you have copy and pasted from `shock_decomposition`. When doing so, you define the anchor of the options a second time.

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+the decomposition. Groups of shocks are defined in @xref{shock_groups} block.
+
+@item colormap = @var{COLORMAP_NAME}
+@anchor{colormap}. Controls the colormap used for the shocks decomposition
+graphs. See @code{colormap} in Matlab/Octave manual.
+
+@item nograph
+@anchor{nograph}. Only shock decompositions are computed and stored in @code{oo_.shock_decomposition},
```
JohannesPfeifer: This should be `oo_.realtime_shock_decomposition`

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+but no plot is made (@xref{plot_shock_decomposition}).
+
+@item presample
+@anchor{presample}. First data point from which recursive realtime shock decompositions are computed, i.e. for T=(presample:nobs).
+
+@item forecast
+@anchor{forecast}. Compute shock decompositions up to T+k periods, i.e. get shock contributions to k-step ahead forecasts.
+@end table
+
```
JohannesPfeifer: I would add an `@outputhead` here (it also seems to be missing for `shock_decomposition`)

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+Field @code{pool} stores the pooled decomposition @xref{plot_shock_decomposition}.
+Fields @code{time_*} store the vintages of realtime shock decompositions.
+
+@vindex oo_.conditional_shock_decomposition
+The results of realtime conditional decompositions are stored in the field @code{oo_.conditional_shock_decomposition}, which is a structure.
+Field @code{pool} stores the pooled decomposition @xref{plot_shock_decomposition}.
+Fields @code{time_*} store the vintages of conditional forecast shock decompositions.
+
+@vindex oo_.realtime_forecast_shock_decomposition
+The results of realtime forecast decompositions are stored in the field @code{oo_.realtime_forecast_shock_decomposition}, which is a structure.
```
JohannesPfeifer: Here we need to expand how exactly the forecasts are stored as there are up to k forecast horizons

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+@anchor{forecast}. Compute shock decompositions up to T+k periods, i.e. get shock contributions to k-step ahead forecasts.
+@end table
+
+@vindex oo_.realtime_shock_decomposition
+The results of realtime decompositions are stored in the field @code{oo_.realtime_shock_decomposition}, which is a structure.
+Field @code{pool} stores the pooled decomposition @xref{plot_shock_decomposition}.
+Fields @code{time_*} store the vintages of realtime shock decompositions.
+
+@vindex oo_.conditional_shock_decomposition
+The results of realtime conditional decompositions are stored in the field @code{oo_.conditional_shock_decomposition}, which is a structure.
```
JohannesPfeifer: I don't understand what the difference between `oo_.realtime_shock_decomposition` and `oo_.conditional_shock_decomposition` is.

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+@code{yoy} for year-on-year plots of growth rates, 
+@code{aoa} for annualized variables, i.e. the value in the last quarter for each year is plotted.
+Default value: @code{empty}, i.e. standard period-on-period plots [qoq for quarterly data].
+
+@item fig_name = @var{FIG_NAME}
+@anchor{fig_name}. Specifies a user-defined keyword to be appended to the default figure name set by @code{plot_shock_decomposition}.
+This can avoid to overwrite plots in case of sequential calls to @code{plot_shock_decomposition}.
+
+@item write_xls 
+@anchor{write_xls}. Saves shock decompotions to excel.
```
JohannesPfeifer: This should read `decomposition`

doc/dynare.texi
```diff
@@ -7126,6 +7130,163 @@ shocks_decomposition(use_shock_groups=group1);
...
+
+@item realtime = @var{INTEGER}
+@anchor{realtime}. Which kind of shock decomposition to plot. @var{INTEGER} can take following values:
```
JohannesPfeifer: This decription of the distinction between types should be introduced much earlier. Also, what defines the "standard" relative to the others?

rattoma:
@JohannesPfeifer thanks a lot. I will go through all your notes. Concerning factorization of realtime, I may proceed directly with that. I principle we could have everything into `shock_decomposition.m` and also the dynare command could still be something like:
`shock_decomposition(realtime=1)`
without the need to have the new one `realtime_shock_decomposition`
Would this be a better option for you?

matlab/shock_decomposition.m
rattoma: you are right. better to to rollback to your previous implementation then. sorry

matlab/utilities/graphics/colorspace.m
```diff
@@ -0,0 +1,515 @@
...
+% ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+% LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+% CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+% SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+% INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+% CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+% ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+% POSSIBILITY OF SUCH DAMAGE.
+
+%%% Input parsing %%%
```
rattoma: tortoise GIT I fear

doc/dynare.texi
rattoma: I thought to have fixed it before committing but did not check the result of the commit ... I'll check.

doc/dynare.texi
rattoma: no. The case 0 is the current behavior of DYNARE which I thought was intentional. The smoothed shock in period one are also used, which means that the initial condition is implicitly the one in period 0. I set this option in order not to change the default behavior of dynare w.r.t. to the past.

doc/dynare.texi
rattoma: OK. this means I have to remove the anchor?

JohannesPfeifer: @rattoma In case the anchor already exists, it seems best to just use e.g. @xref{use_shock_groups} as is already the case for `nobs`

JohannesPfeifer:
@rattoma I would prefer this factorization, but I would like to know what the others think

doc/dynare.texi
JohannesPfeifer: I see. We should definitely document this behavior very clearly.

doc/dynare.texi
rattoma: @JohannesPfeifer actually, what I do is, for each period t=presample:T, to compute historical shock decomposition [1:t]. In principle it is still the historical decomposition, but it is computed at time t, i.e. without observing data in [t+1:T]. How do we name it to avoid confusion? Just realtime, or realtime-historical?

doc/dynare.texi
rattoma: realtime shock decomposition sets the initial condition for the shock decomposition in period 1 (or 0). So it computes the contribution of shocks back to the past, i.e. in (1:t)
conditional shock decomposition, sets the initial condition in t-1, so only computes the effect of one period shock give the state in t-1, i.e. it is just a 1-period shock decomposition from t-1 to t. In practice I decompose the update step of the Kalman filter.

doc/dynare.texi
rattoma: the standard is the usual one, computed by `shock_decomposition`

matlab/realtime_shock_decomposition.m
rattoma: I got crashes related to this. 
I think that, in principle, if `evaluate_smoother` was run earlier with full sample it should not be needed. 

However, if the first time we run `evaluate_smoother` it is for realtime shock decomposition, then the persistent dataset_ initialized would be shortest [`options_.nobs=j`] and we will get a crash in the next iteration of the realtime decompositions because sample is too short in dataset_

By the way, I think we have to add `M_`  as output argument of `evaluate_smoother` to ensure consistency between oo_ and M_ when parameter_set changes params w.r.t. the M_ in input.

doc/dynare.texi
JohannesPfeifer: @rattoma realtime historical makes sense to me

matlab/realtime_shock_decomposition.m
JohannesPfeifer: @rattoma 
1. I see. I think we should aim at properly setting all variables for calls to functions instead of relying on something being empty. Also, I would prefer getting rid of the persistent variable to a clear statement. Your `clear` solves the problem only for `realtime_shock_decomposition`, but it is most likely going to occur in other contexts as well
2. That depends on your aim. I prefer not to have to keep track of whether commands changed the parameter set. The current behavior assures that `shock_decomposition` does not have any side effects on other functions.

doc/dynare.texi
JohannesPfeifer: @rattoma I see. Then I would add this information to the manual

doc/dynare.texi
JohannesPfeifer: @rattoma I meant that standard/0 should be something like 

> historical shock decomposition of Y(t|T) for t=1:T

doc/dynare.texi
JohannesPfeifer: @rattoma I meant that standard/0 should be documented to be something like 

> historical shock decomposition of Y(t|T) for t=1:T

matlab/realtime_shock_decomposition.m
rattoma: I agree, but I would propose to add M_ as optional last output argument, in case one wants to store all structures in a consistent way after calling `evaluate_smoother`

rattoma:
@JohannesPfeifer I made a first bunch of commits following your notes.

matlab/realtime_shock_decomposition.m
JohannesPfeifer: @rattoma But we currently only save the results in `oo_` and do not alter `options_`  or `M_`

rattoma: Indeed, but assume I want to save a mat file like `<mod>_results.mat`: there I would like to save oo_ and the associated M_. Moreover, `options_` and `bayestopt_` are already optional output args of evaluate_smoother.

rattoma: I am referring to the use of evaluate_smoother in general, not to the specific call in shock decomposition [here it is OK I think].

JohannesPfeifer: @rattoma Do you know whether the default behavior in 4.4.3 was to alter `M_.params`? It might be the case that I erroneously changed the old behavior in 2f717b5adc5a87f663c5c080f2963d1f65d1933e to not save the parameters. In that case, `M_` should regularly be outputted.

rattoma: As far as `evaluate_smoother` is concerned, I think `M_` was updated via `global M_` in `DsgeSmoother` and `evaluate_smoother`. Again, this should not be an issue for shock decomposition

JohannesPfeifer: @rattoma That confirms my suspicion. But then what needs to be done is output `M_` in `evaluate_smoother` and make sure that callers like `shock_decomposition` pass changes in `M_.params` back to the base workspace.

MichelJuillard:
M_.params can be modified by at least the steady state function, maybe also by dsge_var

JohannesPfeifer:
@MichelJuillard There are various functions that alter `M_.params`, most importantly `dynare_estimation_1` so that after estimation one can call `stoch_simul` at the new parameter set. The question here is whether after a call like `shock_decomposition(parameter_set='prior_mean')`, the parameters in `M_.params` should now be at the prior mean or at whatever it was before.

MichelJuillard:
I think it would be better is only estimation() reset M_.params in that way. We should probably provide a utility that sets M_.params from various options before using it for some Matlab computations in the *.mod file

rattoma:
I agree we should not necessarily force to update M_ after using, e.g., shock decomposition. To ensure M_ consistent with oo_, one can first set M_ and later call
`shock_decomposition(parameter_set='calibration')`
However, I think evaluate_smoother should give back M_ as output: whether this will be updated in the base workspace is another story, but we should allow to retrieve the result of an update of M_.params if desired.

It is also true that 2f717b5 changed the behaviour of any function calling the smoother. Before that, M_ was always implicitly updated in the base WS. Users might not expect/understand this behavior is changed, so this needs to be clearly acknowledged. 

However, if the idea about globals was to remove them while preserving the old behavior regarding the status of variables in the base workspace, then we should update M_ in any case ...

MichelJuillard:
@rattoma you have a point about backward compatibility

JohannesPfeifer:
@rattoma @MichelJuillard I moved the last discussion to a new issue #1373. Regardless of the decision, returning `M_` as a function output of `evaluate_smoother` (as @rattoma proposes) is benign and more flexible and therefore should be done. Whether one passes this argument back is then a separate matter

MichelJuillard:
OK for me

rattoma:
I made another couple of fixes and added a test example to my PR. This still misses the pre-processor for  plot_shock_decomposition and realtime_shock_decomposition.

The PR can be tested in my view.

rattoma:
@JohannesPfeifer 
Regarding  #1373, does it make sense that, after your PR  	b191600 on M_ given back by evaluate smoother, shock_decompositions gives back `M_` and `oo_.shock_decomposition` but `oo_.SmoothedVariables` etc. which are computed internally by evaluate_smoother not? shouldn't we use `oo_` in output argument of the call evaluate_smoother in place of `oo` ?

In this case we would have the whole chain `M_` - `oo_` fully consistent?

JohannesPfeifer:
@rattoma I would prefer for `shock_decomposition` to only return `oo_.shock_decomposition`, because that is what the command is supposed to do. If we follow your logic, we should update all objects in `oo_`, like e.g. `oo_.dr` at the new parameter set.

rattoma:
@JohannesPfeifer OK for me

rattoma:
although just capturing the oo output to the call to evaluate_smoother in shock_decomposition would update automatically everything

JohannesPfeifer:
@rattoma I just see that this was default behavior before my change. I will adust the PR accordingly

tests/shock_decomposition/ls2003_plot.mod
```diff
@@ -0,0 +1,144 @@
...
+rho_ys ,beta_pdf,0.8,0.1;
+rho_pies,beta_pdf,0.7,0.15;
+stderr e_R,inv_gamma_pdf,1.2533,0.6551;
+stderr e_q,inv_gamma_pdf,2.5066,1.3103;
+stderr e_A,inv_gamma_pdf,1.2533,0.6551;
+stderr e_ys,inv_gamma_pdf,1.2533,0.6551;
+stderr e_pies,inv_gamma_pdf,1.88,0.9827;
+end;
+
+estimation(datafile=data_ca1,first_obs=8,nobs=79,mh_nblocks=10,prefilter=1,mh_jscale=0.5,mh_replic=0);
```
JohannesPfeifer: @rattoma This should be `datafile='../ls2003/data_ca1'`

JohannesPfeifer:
@rattoma Sorry for keeping you busy. Another round of comments:

1. I am having trouble with this PR. In `graph_decomp.m` in the line `~isempty(opts_decomp.fig_mode)` I get the the field does not exist. I guess you are missing a 
`options_.shock_decomp.fig_mode='';` 
in `global_initialization`. 

2. Could you move the storing of figure output to a subfolder, like `[M_.fname, filesep, 'graphs']`? With the new commands, there are really many files created.

3. The `realtime_shock_decomposition.m` graphs overwrite the ones of `shock_decomposition` (and have the same Matlab figure name). Is that desired? Also, it seems `plot_shock_decomposition.m` behaves differently for realtime by passing `fig_names =_realtime_pool`, which assures a unique name.

4. Would it be possible to provide more expressive figure titles? Currently you use the same `fig_names` for `dyn_saveas` and `dyn_figure(DynareOptions,'Name',[`. I would prefer an approach as in `plot_identification.m`.

5. In the unit test, could you please add
`options_.TeX=1` before the estimation command and

```
collect_LaTeX_Files;
if system(['pdflatex -halt-on-error -interaction=batchmode ' M_.fname '_TeX_binder.tex'])
    error('TeX-File did not compile.')
end 
```

at the end of the unit test.

6. If you look at the created TeX-file, you will see that i) all TeX-loaders have the figure caption `Historical shock decomposition`, ii) the label is always `\label{Fig:shock_decomp:`, and iii) for groups like `ls2003_plot_shock_decomposition_yoy_R_obs_MR_group_trade_detail` the caption is missing information on the groups and the tranformation (like yoy)

7. The detailed graphs created by `graph_decomp_detail` would greatly benefit from a legend in the first subplot. What are the black and grey bars? The black line is the data, I guess?

rattoma:
@JohannesPfeifer thanks a lot. this is very useful. I made commits to the PR which should fix all your problems 1 to 6.
1. I removed the useless internal option `fig_mode` relying on the already erxisting `type`
2. done
3. I modified the logic: `realtime_shock_decomposition` does NOT plot anything while `shock_decomposition` now relies on `plot_shock_decomposition` without calling `graph_decomp` directly.
4. done
5. done
6. done
7. black bar is the shock or shock group under focus. grey is all the other contrib that add up to the data. black line is the data. Since there is no legend for the data in the usual colored plots, I would only make a caption for black/grey bars.

rattoma:
7. also fixed

JohannesPfeifer:
@rattoma Thanks. This is a lot better. One minor quibble. The captions are still not unique. In the compiled TeX-file, we have for example 10 figures with the caption
``Historical shock decomposition: dq`` 
What distinguishes those plots?

rattoma:
@JohannesPfeifer indeed they start with the same name, but there should be extra text appended after the name of the variable, right? Or I only see extended captions on my pc?
By the way, I cannot run pdflatex on my pc for TeX_binder, so I compile with the steps latex, dvips, ps2pdf. Not sure whether this makes the difference for captions.
In any case, the preamble of captions can be better differentiated.

JohannesPfeifer:
@rattoma Sorry for the noise. It still had the graphs from a previous run in the folder. Everything is fine now.

rattoma:
@JohannesPfeifer thank you. I still made some adjustments to names/captions.

rattoma:
@JohannesPfeifer 
I committed an hopefully last batch of changes/adds. From now on I should be able to commit myself for a freeze of this branch and allow only for bug fixes to the current behavior.
many thanks

matlab/utilities/dataset/lagged.m
```diff
@@ -0,0 +1,32 @@
+function xlag = lagged(x, n);
```
JohannesPfeifer: @rattoma I wonder whether it would make more sense to add a `lagmatrix`-function to the `missing`-folder, because we are essentially duplicating the functionality of that function. See also `http://www.spatial-econometrics.com/arfima/lagmatrix.m`

matlab/annualized_shock_decomposition.m
```diff
@@ -0,0 +1,195 @@
+function [z, endo_names, endo_names_tex, steady_state, i_var, oo_] = annualized_shock_decomposition(oo_, M_, opts, i_var, t0, t1, realtime_, vintage_, steady_state,GYTREND0,var_type,islog)
+% function oo_ = annualized_shock_decomposition(oo_,t0,options_.nobs);
+% Computes annualized shocks contribution to a simulated trajectory. The fields set are
```
JohannesPfeifer: @rattoma We still need to document the respective fields in `oo_` in the manual. It would also be great if you could clarify whether this simply annualizes growth rates (*4 as is customary in US NIPA tables) or whether we are talking about year on year comparisons.

matlab/utilities/dataset/quarterly2annual.m
```diff
@@ -0,0 +1,79 @@
...
+    yss=exp(yss);
+    y=y-yss;
+end
+switch type
+    case 1
+        yass = yss*(exp(-GYTREND0*3)+exp(-GYTREND0*2)+exp(-GYTREND0)+1);
+        tmp = lagged(y,3)*exp(-GYTREND0*3)+lagged(y,2)*exp(-GYTREND0*2)+lagged(y,1)*exp(-GYTREND0)+y; % annualized level
+    case 2
+        yass = yss*(exp(-GYTREND0*3)+exp(-GYTREND0*2)+exp(-GYTREND0)+1)/4;
+        tmp = (lagged(y,3)*exp(-GYTREND0*3)+lagged(y,2)*exp(-GYTREND0*2)+lagged(y,1)*exp(-GYTREND0)+y)/4; % annualized level
```
JohannesPfeifer: In case of a deflator, shouldn't this be a geometric average instead of an arithmetic one?

rattoma: In this case I am deriving the price level, which I think is the arithmetic average of the quarterly price levels. Geometric average should apply to growth factors?
Here I also correct for constant growth in the case we are detrending the model with `trend_var`. By the way, it would be useful to store info about trends and deflators of detrended models into `M_`.

matlab/annualized_shock_decomposition.m
rattoma: It is a bit more complicated since I generalize to the case where the variables are detrended, so I correct for the growth factor.

matlab/utilities/dataset/lagged.m
rattoma: Could be done. However, not sure whether I replicate all the functionality you mention [e.g provide a vector of lags in input]. Also, this is not an  'official' matlab toolbox?

matlab/annualized_shock_decomposition.m
rattoma: moreover I sum the quarterly flows, not really multiply by 4.

JohannesPfeifer: 1. But summing up the flows is a year on year comparison
2. How do you deal with stocks? Or is that the "deflator" case?

matlab/utilities/dataset/quarterly2annual.m
JohannesPfeifer: @rattoma This seems to apply to stocks and price levels. But shouldn't we also have a case for geometric averages which apply to growth rates and interest rates? Regarding the storage to deflators, I will add a comment to #1162

matlab/utilities/dataset/lagged.m
JohannesPfeifer: @rattoma I guess we should leave it for now. A `lagmatrix` command has some years ago become part of Matlab's econometrics toolbox, but before it was only in LeSage's toolbox.

matlab/utilities/dataset/quarterly2annual.m
rattoma: @JohannesPfeifer case 1 is flow, case 2 is deflator, case 3 is stock. 
One remaining issue is that we need also to construct annual data that preserves consistency between nominal variables, prices and volumes. So, I will add a case where we give as input twe series [one for volumes and one for associated price index], and construct annual nominal flow, annual real flow and the annual deflator by difference [i.e. annual price is the volume weighted aggregation of quarterly prices].

Concerning growth rates and interest rates with geometric average: what we are aiming here is the construction of true annual data, which is different to the YoY concept. So, the annual growth rate that I am considering is derived from the reconstructed annual levels and hence from the AoA growth rate thereof.

I think what you are discussing may be an extension of the YoY case.

JohannesPfeifer: @rattoma I see. 
1. But why is case 3 a stock? If you observe a capital stock at 4 quarters in a year, what is the yearly value? I thought it is the arithmetic average over the quarterly values as would be the case with deflators.
2. Similarly, if you observe the average effective federal funds rate for four quarters, the annual value for the calender year should be the geometric average of the quarterly gross interest rates.

rattoma: concerning stocks, in most cases it is the final value of the flows feeding the stock over the year. For example Govt. stock of debt annual is the debt at the end of the year. I think one notable exception to this notation is the stock of population which I think is the average as you say. As far as I know, most stock series have the notation of 'end-of-period'. 

We may not get to an end in this q2a utility for the general user... Perhaps we may think of letting this not fully documented/fully committed to address all possible cases?

We may ' rename' cases 1 2 3 as:
1=sum
2=average
3=end-of-period

and the user can map his own situation to what is available

perhaps we may add
4=geometric average

without taking any stance/commitment to produce correct annual series for stocks flows deflators etc ...

JohannesPfeifer: @rattoma That seems like a good compromise, particularly as the averaging of interest rates differs for net and gross rates.

rattoma:
@JohannesPfeifer done for annual aggregation types

rattoma:
@JohannesPfeifer I added several fixes, 1 missing file and extensions to annual aggregator.

rattoma:
@JohannesPfeifer I added a commit to this PR that tries to fix a number of issued related to the persistent dataset_ variable.

rattoma:
@JohannesPfeifer I modified my commit, also applying the patch to evaluate_likelihood which might display similar problems with qz_criterium.

stepan-a:
Manually merged.

PR #1374, opened 2017-01-09
evaluate_smoother.m: pass updated M_.params back to base workspace

### LINKED ISSUE #1373: Decide on whether evalute_smoother (and others) should set M_.params
This picks up the discussion in https://github.com/DynareTeam/dynare/pull/1372#issuecomment-271336355

I agree with @MichelJuillard that changing `M_.params` in various functions is a bad idea as it makes it hard to trace what was going on in the mod-file. At the same time, I agree with @rattoma that keeping `M_.params` unchanged when calling `evaluate_smoother` would imply a break with previous versions and therefore mean a loss in backward compatibility. I could live with that, but usually @MichelJuillard prefers to be cautious in the regard. If we want to preserve backward compatibility, we need to make sure that `shock_decomposition` returns `M_` from `evaluate_smoother` to the base workspace, which was broken by https://github.com/DynareTeam/dynare/commit/2f717b5adc5a87f663c5c080f2963d1f65d1933e

JohannesPfeifer:
Closes #1373

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -7042,7 +7042,10 @@ Specify the parameter set to use for running the smoother. The
 @var{PARAMETER_SET} can take one of the following seven values:
 @code{calibration}, @code{prior_mode}, @code{prior_mean},
 @code{posterior_mode}, @code{posterior_mean},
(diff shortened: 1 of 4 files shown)


stepan-a:
Manually merged.

PR #1375, opened 2017-01-10
storage of kstep ahead filter covariance and variables

rattoma:
- store oo_.FilteredVariablesKStepAheadVariances only if options_.filter_covariance. this can save a lot of memory for large models

- oo_.FilteredVariables must be stored independently to a previous MCMC. evaluate_smoother must in any case provide full info in oo_

- in prior_posterior_statistics options_.filtered_vars does not always imply filtered variables are computed. options_.nk is more robust.


JohannesPfeifer:
@rattoma 
1. If I understand it correctly, storing `FilteredVariablesKStepAheadVariances` regardless of whether it had been requested was a bug in `4.4.3`. If you agree, we should add this to the bug list.
2. Yes, that was my mistake in the transition to `store_smoother_results`
3. In which cases can it happen that `prior_posterior_statistics_options_.filtered_vars` does not imply filtered variables are computed? This looks like a bug to me that needs to be fixed rather than relying on `options_.nk`. Related to this, it seems we need to clarify the distinction/relation between `filter_step_ahead`, `filter_covariance`, and `filtered_vars`. If I read the manual correctly, `filtered_vars` is supposed to only govern the one-step ahead filtered variables, while `filter_step_ahead` governs `k`-step ahead forecasts. `filter_covariance` seems to tell Dynare to save `FilteredVariablesKStepAheadVariances`, regardless of whether the forecasts were one or `k`-step (here, the manual needs to state that `filter_step_ahead` only saves  `FilteredVariablesKStepAheadVariances` if `filter_covariance` was specified)

rattoma:
@JohannesPfeifer 
1. yes I think this can be seen as a bug
2. yes
3. it will happen when `filter_step_ahead` is declared without declaring `filtered_vars`. This case is handy because if you write `filter_step_ahead=[1 4]` 1-step is already in, so `filtered_vars` is redundant.
In `dynare_estimation_init`, `filter_step_ahead` is set when `filtered_vars` is declared. But not viceversa.
`options_.nk` is in any case set, which triggers my proposal in PR.

The alternative to my PR would be to add the following lines in dynare_estimation_init.m, around line 144:

```
if  isequal(options_.filtered_vars,0) && ~isempty(options_.filter_step_ahead)
    options_.filtered_vars = 1;
end
```

JohannesPfeifer:
I think we need to sort this out more systematically. 
According to the manual:

1. If `filtered_vars` then we set `oo_.FilteredVariables`. If also `filter_covariance`, we set `oo_.FilterCovariance` (Metropolis) or `oo_.Smoother.Variance` (other)
2. If `filter_step_ahead` we set `oo_.FilteredVariablesKStepAhead`


a) What you now introduce is that if `filter_step_ahead` we only set `oo_.FilteredVariablesKStepAhead` if `filter_covariance`. I would be fine with this change, but it must be documented.
b) The current behavior of Dynare deviates from the outlined behavior above. Whenever `filter_step_ahead` is present, we save `oo_.FilteredVariables`, even if `filter_step_ahead` does not contain horizon 1. This strikes me as undesirable behavior and should most probably changed
c) Additionally, we now always save `oo_.Smoother.Variance`, not only when Metropolis was used. This behavior changed from 4.4.3 when we moved to `store_smoother_results`. We might want to change it back

rattoma:
@JohannesPfeifer 
1. My PR seems to uncover my misreanding on filtered_vars. According to the manual, behavior is quite clean, and I would have to set `filtered_vars` to fix it in my dynare files.

a. I would keep this new behavior, documenting it
b. we might just revert my proposal, and I would have to remember to set both `filter_step_ahead` and `filtered_vars` in estimation
c. fine with me

JohannesPfeifer:
@rattoma I don't fully understand your point regarding b). I was saying that the current Dynare behavior (unrelated to the current state of the PR) is not fully consistent with the manual and we may want to change that. @MichelJuillard What do you think?

MichelJuillard:
@JohannesPfeifer I can't look into it for the time being. Sorry

JohannesPfeifer:
@MichelJuillard Let me pose the decisions as a simple question so you don't need to read the whole thread. Should we fill `oo_.FilteredVariables` when a user has not requested `filtered_variables`, but only `filter_step_ahead=2`? We currently do so, but this seems to be against the documented functionality of the `filter_step_ahead`option, which is to only fill `oo_.FilteredVariablesKStepAhead`. Getting rid of this filling of the field would be cleaner, but involves a break with undocumented functionality and therefore backward compatibility. If you prefer compatibility, we would just update the manual to document the current behavior.

MichelJuillard:
Thanks @JohannesPfeifer. Just update the manual

rattoma:
@JohannesPfeifer @MichelJuillard thanks a lot. I will update the PR accordingly

rattoma:
@JohannesPfeifer I updated the manual. Concerning the use of `options_.nk` in `prior_posterior_statistics` this seems consistent with what already coded in `store_smoother_results`. Shall we keep this formulation then?

JohannesPfeifer:
@rattoma For better transparency, I would go for 
``` 
if  isequal(options_.filtered_vars,0) && ~isempty(options_.filter_step_ahead)
    options_.filtered_vars = 1;
end
```
in `dynare_estimation_init.m`

rattoma:
done

stepan-a:
Merged manually.

PR #1378, opened 2017-01-17
Slice fixes

rattoma:
fix bugs with load_mh_file of slice sampler [invhess missing]
maximize sub draws for very short chains

### DIFF
--- a/matlab/dynare_estimation_init.m
+++ b/matlab/dynare_estimation_init.m
@@ -603,7 +603,7 @@ if options_.load_results_after_load_mh
     end
 end
 
-if options_.mh_replic
+if options_.mh_replic || options_.load_mh_file,
     [current_options, options_] = check_posterior_sampler_options([], options_, bounds);
     options_.posterior_sampler_options.current_options = current_options;
 end
--- a/matlab/metropolis_draw.m
+++ b/matlab/metropolis_draw.m
@@ -74,7 +74,7 @@ if init
     else
         if options_.sub_draws>NumberOfDraws*mh_nblck
             skipline()
-            disp(['Estimation::mcmc: The value of option sub_draws (' num2str(options_.sub_draws) ') is greater than the number of available draws in the MCMC (' num2str(NumberOfDraws) ')!'])
+            disp(['Estimation::mcmc: The value of option sub_draws (' num2str(options_.sub_draws) ') is greater than the number of available draws in the MCMC (' num2str(NumberOfDraws*mh_nblck) ')!'])
             disp('Estimation::mcmc: You can either change the value of sub_draws, reduce the value of mh_drop, or run another mcmc (with the load_mh_file option).')
             skipline()
             xparams = 1; % xparams is interpreted as an error flag
--- a/matlab/metropolis_draw.m
+++ b/matlab/metropolis_draw.m
@@ -70,9 +70,9 @@ if init
     mh_nblck = options_.mh_nblck;
     % set sub_draws option if empty
     if isempty(options_.sub_draws)
-        options_.sub_draws = min(options_.posterior_max_subsample_draws, round(.25*NumberOfDraws));
+        options_.sub_draws = min(options_.posterior_max_subsample_draws, round(NumberOfDraws*mh_nblck));
     else
-        if options_.sub_draws>NumberOfDraws
+        if options_.sub_draws>NumberOfDraws*mh_nblck
             skipline()
             disp(['Estimation::mcmc: The value of option sub_draws (' num2str(options_.sub_draws) ') is greater than the number of available draws in the MCMC (' num2str(NumberOfDraws) ')!'])
             disp('Estimation::mcmc: You can either change the value of sub_draws, reduce the value of mh_drop, or run another mcmc (with the load_mh_file option).')
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -5264,8 +5264,8 @@ smoothed shocks,  forecast, moments,  IRF).  The  draws used  to compute
 these posterior moments are sampled uniformly in the estimated empirical
 posterior  distribution (@i{ie}  draws of  the MCMC).   @code{sub_draws}
 should  be  smaller than  the  total  number  of MCMC  draws  available.
-Default:  @code{min(posterior_max_subsample_draws,0.25*Total  number  of
-draws)}
+Default:  @code{min(posterior_max_subsample_draws,(Total  number  of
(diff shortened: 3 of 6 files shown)


matlab/dynare_estimation_init.m
```diff
@@ -603,7 +603,7 @@
     end
 end
 
-if options_.mh_replic
+if options_.mh_replic || options_.load_mh_file,
```
JohannesPfeifer: @rattoma Should we add a test case for this? And do you know whether mh_recover works with slice?

matlab/metropolis_draw.m
```diff
@@ -70,9 +70,9 @@
     mh_nblck = options_.mh_nblck;
     % set sub_draws option if empty
     if isempty(options_.sub_draws)
-        options_.sub_draws = min(options_.posterior_max_subsample_draws, round(.25*NumberOfDraws));
+        options_.sub_draws = min(options_.posterior_max_subsample_draws, round(NumberOfDraws*mh_nblck));
```
JohannesPfeifer: @rattoma I am fine with this change in the default behavior, but the manual on `sub_draws` needs to be updated accordingly

matlab/dynare_estimation_init.m
rattoma: @JohannesPfeifer 
1. I think about it. If I remember well, without my fix, dynare would not crash but would exit doing nothing.
2. admittedly, I never really tried `mh_recover`, even with metropolis. Always did it by hand ... should be done ...

matlab/metropolis_draw.m
rattoma: @JohannesPfeifer I updated my PR accordingly

matlab/dynare_estimation_init.m
JohannesPfeifer: @rattoma 
1. It should be straightforward to adjust https://github.com/DynareTeam/dynare/blob/master/tests/estimation/MH_recover/fs2000_recover_tarb.mod to test Slice. 
2. Similarly, the `load_mh_file`-related parts at the bottom of https://github.com/DynareTeam/dynare/blob/master/tests/estimation/fs2000.mod could be ported to https://github.com/DynareTeam/dynare/blob/master/tests/estimation/slice/fs2000_slice.mod

rattoma:
@JohannesPfeifer the trap you mention should be already in the initialization of metropolis_draw, line 75 onwards:
```
        if options_.sub_draws>NumberOfDraws*mh_nblck
            skipline()
            ...
```
So, we should never have B>NumberOfDraws*mh_nblck at that point.
Still, we may put a counter there to ensure in any case not too many iterations.

JohannesPfeifer:
@rattoma You are right, no additional check is necessary. But checking identical draws via identical log-likelihoods strikes me as error-prone. Take for example a symmetric distribution. We would say both sides of the distribution are identical. Also, it might make more sense to rely on https://www.mathworks.com/help/stats/randsample.html to generate a sample of draws without replacement than sampling and rejecting as we currently do as many draws may be rejected if we draw close to the number of available draws as subdraws

stepan-a:
@rattoma I have a problem with f70f45d. The testsuite enters in infinite loops probably because of your attempt to avoid repeated draws. If `mh_replic` is small enough and the acceptance rate close to one quarter, it may be impossible to find `options_.sub_draws` distinct draws... Actually, I do not understand why you want to avoid repetitions. These repetitions are key property for the metropolis algorithm to converge to the posterior distribution. So I don't think that what you try to achieve here is correct. Also when `B==NumberOfDraws*mh_nblck`, if my understanding is correct, you do not impose the uniqueness  of the draws (you take them as they are, with repetitions).

I try to revert to run the testsuite by partially reverting this commit. If it works (for now the PR does not work because of the infinite loops). I merge into master. We can discuss this issue later...

stepan-a:
Manually merged. @rattoma If you are not happy with my partial revert of f70f45d (see  e71ed1b) please open a new issue.

JohannesPfeifer:
@rattoma I agree with @stepan-a. I missed that the condition you imposed is based on the likelihood and thus rules out  MCMC draws where the proposed draw was rejected. Discarding them from the posterior subdraws will make the draws violate the detailed balance condition, implying that they are not draws from the actual posterior.

rattoma:
@JohannesPfeifer @stepan-a you are right. The point was that with the slice sampler there are never repeated draws in the posterior so that loop avoids repeated sampling of the same 'line' of the posterior MC matrix of the sampler.
Indeed I did not think that with metropolis the problem is different: we need to keep the properties of the balance condition provided by the acceptance/rejection algorithm.

I would add, however, a PR where I keep the reverted trap only for slice.

Concerning the metropolis, I think that we may also use a similar condition, but using the 'row' index of the draw in the posterior MC sample and not the likelihood.

let me provide an example:
Assume a metropolis draw x0 which is not replaced by a new accepted draw for indices: `[j1 j2 j3 j4 j5 j6]`.

Then, should we only accept subdraws that takes a sequence `[j2 j3 j6]` but avoid sequences like `[j2 j2 j6 j6]` ? 

The latter would be something like a bootstrap re-sampling, which I do not dislike but can also have disadvantages, e.g. it may violate the balance condition the other way round, i.e. by oversampling x0 [especially for small sample sizes?]. 

Or perhaps better, could we add a subdraw option to allow or not bootstrap re-sampling? The current approach is implicitly a bootstrap re-sampling algorithm [which would be the default]. We may make this more explicit with the option and allow to offset it if desired.

what do you think?

JohannesPfeifer:
@rattoma My intuition was the following: we get a large set of draws from the posterior. Instead of computing posterior statistics for all of the available posterior draws, we use an i.i.d sample of size `sub_draws` from the posterior distribution by sampling with replacement. 
What you suggest is sampling without replacement. This would imply that the subsample does not satisfy "independence" anymore.

rattoma:
@JohannesPfeifer: you are right. 

then, should we also allow `B>NumberOfDraws*mh_nblck`? It seems to me that bootstrapping literature suggests to take a bootstrap sample as large as possible, beyond the possible small empirical sample available.

stepan-a:
@rattoma  I agree, in principle we could  allow for `B>NumberOfDraws*mh_nblck`, but in practice we choose a smaller `B` for saving time (otherwise we would always take all the draws from the MCMC, except from the burn-in period, by default).

JohannesPfeifer:
@rattoma I would also keep the restriction as a warning for naive users

rattoma:
Agree

PR #1379, opened 2017-01-18
Adjust Octave unit tests

JohannesPfeifer:
@houtanb I also looked at the other failing tests. For the `quantile.m` tests I am not able to replicate the issue. I guess one needs to check on the machine running the tests.

For the `dseries` tests, the problem seems to be that Octave's `io` package cannot handle `xls` files unless Excel is installed. An easy workaround is saving the four `http://www.dynare.org/Datasets/dseries/dynseries_test_data-*.xls` files in `xlsx` format and then adjusting the `dseries.m` tests 9 to 12 to load those

### DIFF
--- a/matlab/distributions/lpdfgweibull.m
+++ b/matlab/distributions/lpdfgweibull.m
@@ -279,7 +279,7 @@ end
 %$ 
 %$ try
 %$    if isoctave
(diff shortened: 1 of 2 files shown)


houtanb:
@JohannesPfeifer thanks for your help

PR #1382, opened 2017-01-25
Error msg undeclared model vars

### LINKED ISSUE #1286: issue undeclared model variable errors at the end of model parsing
Currently, when a variable is in an equation but not declared, the preprocessor issues an error and stops parsing. This is a problem during model development as a user can potentially need to run dynare several times before catching all undeclared variables. To fix this, `preprocessor/ParsingDriver.cc` needs to be modified to check the existence of variables at the end of the `model` block.

houtanb:
Closes #1286


stepan-a:
Manually merged.

PR #1383, opened 2017-01-26
readme.md: Fix typos and add info on make clean

JohannesPfeifer:


README.md
```diff
@@ -194,6 +194,13 @@ or
 ```
 make deterministic_simulations/lbj/rbc.o.drs
 ```
+Note that only tests will be executed where the `m.trs/o.trs` does not yet exist. You can run
+```
+make clean
+```
+in the `tests` folder to delete all files that were not originally part of this folder. However, be careful when doing so, 
+because this command will also delete all files you may have intentionally copied to the folder. In that case, you may instead
```
houtanb: This is not true. `make clean` deletes files following the rules in Makefile.am. If you copy, for example, a `.txt`, `.m`, `.mod`, or many other types of files to a folder, those files will not be deleted. You can see the rules in `tests/Makefile.am` under `clean-local`

JohannesPfeifer:
@houtanb Thanks. I updated the commit.

PR #1385, opened 2017-01-30
- make info on inf/nan/complex more informative, which in many cases …

rattoma:
…allows to point to the block that creates problems

- allow for relaxation steps for difficult problems [on top of robust solver ... if the case]

### DIFF
--- a/matlab/perfect-foresight-models/sim1.m
+++ b/matlab/perfect-foresight-models/sim1.m
@@ -173,33 +173,9 @@ for iter = 1:options.simul.maxit
         % relaxation step, reducing length of update of previous
         % iteration
         if verbose
-            dyy=reshape(dy,[size(endogenousvariables,1) periods])';
-            if any(isnan(dy))
-                indx = find(any(isnan(dyy)));
-                endo_names=cellstr(M.endo_names(indx,:));
-                disp('Current iteration provided NaN for variables:')
-                fprintf('%s, ',endo_names{:}),
-                fprintf('\n'),
-                nan_flag = 1;
-                %         keyboard;
-            end
-            if any(isinf(dy))
-                indx = find(any(isinf(dyy)));
-                endo_names=cellstr(M.endo_names(indx,:));
-                disp('Current iteration diverged (Inf) for variables:')
-                fprintf('%s, ',endo_names{:}),
-                fprintf('\n'),
-            end
-            if any(~isreal(dy))
-                indx = find(any(~isreal(dyy)));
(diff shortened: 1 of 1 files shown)


matlab/perfect-foresight-models/sim1.m
```diff
@@ -163,8 +168,45 @@
...
+        % iteration
+        if verbose
+            dyy=reshape(dy,[size(endogenousvariables,1) periods])';
+            if any(isnan(dy))
+                indx = find(any(isnan(dyy)));
+                endo_names=cellstr(M.endo_names(indx,:));
+                disp('Current iteration provided NaN for variables:')
+                fprintf('%s, ',endo_names{:}),
+                fprintf('\n'),
+                nan_flag = 1;
```
JohannesPfeifer: I think you should remove the commented out `keyboard` as well as the setting of the unused `nan_flag` here and below

matlab/perfect-foresight-models/sim1.m
```diff
@@ -163,8 +168,45 @@
             dy = -lin_solve_robust( A, res, verbose );            
         else
             dy = -lin_solve( A, res, verbose );
+    end
+    if iter>1 && (any(~isreal(dy)) || any(isnan(dy)) || any(isinf(dy)) || (err/err0)>10),
+        % relaxation step, reducing length of update of previous
+        % iteration
+        if verbose
```
JohannesPfeifer: It seems that this display of debugging info is implemented twice here. An anonymous function might be a better way to factorize the code.

matlab/perfect-foresight-models/sim1.m
```diff
@@ -163,8 +168,45 @@
...
+            if any(~isreal(dy))
+                indx = find(any(~isreal(dyy)));
+                endo_names=cellstr(M.endo_names(indx,:));
+                disp('Current iteration provided complex number for variables:')
+                fprintf('%s, ',endo_names{:}),
+                fprintf('\n'),
+                
+            end
         end
+        disp('Try relazation step.')
```
JohannesPfeifer: This should be "relaxation"

rattoma:
@JohannesPfeifer thanks a lot. this is done now.

stepan-a:
After discussion with @rattoma, we agreed to abandon the part of the patch related to the relaxation step. This feature is already available by using other solvers (with `stack_solve_algo=7`) where algorithms with line search are available. I will merge the rest of the patch manually.

stepan-a:
Manually merged (only the section concerning the displayed informations when NaN/Inf or complex numbers occur).

PR #1388, opened 2017-02-03
Gsa bug-fixes and fix #1016

### LINKED ISSUE #1016: Clean up setting of options in dynare_sensitivity.m
Currently, it is very intransparent and prone to bugs. 
-  `neighborhood_width` forces 

```
if options_gsa.neighborhood_width,
    options_gsa.pprior=0;
    options_gsa.ppost=0;
end
```

but the manual says that  `neighborhood_width` only works when `pprior=0` and `ppost=0`. This seems like a bug. If we reset options set by the user, we should consistently provide warnings on the screen. 
- Even more problematic, without any indication in the manual, using `morris=1` overwrites almost all other user-specified options:

```
if options_gsa.morris==1,
    if ~options_gsa.identification,
        options_gsa.redform=1;
    end
    options_gsa.pprior=1;
    options_gsa.ppost=0;
    %options_gsa.stab=1;
    options_gsa.glue=0;
    options_gsa.rmse=0;
    options_gsa.load_rmse=0;
    options_gsa.alpha2_stab=1;
    options_gsa.ksstat=1;
    options_gsa.pvalue_ks=0;
    options_gsa.pvalue_corr=0;
    OutputDirectoryName = CheckPath('gsa/screen',M_.dname);
else
    OutputDirectoryName = CheckPath('gsa',M_.dname);
end
```

In particular, that results in none of the user-requested output being displayed. See http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=7115. @rattoma Is there a particular reason for this? I could not find any apparent incompatibility of `morris=1` with e.g. posterior sampling.

rattoma:


matlab/gsa/scatter_plots.m
```diff
@@ -0,0 +1,190 @@
...
+                % item3 = uimenu(hcmenu,'Label','solid','Callback',hcb3);
+                % Locate line objects
+                hlines = findall(h,'Type','line');
+                % Attach the context menu to each line
+                for line = 1:length(hlines)
+                    set(hlines(line),'uicontextmenu',hcmenu)
+                end
+            end
+            %%
+
```
JohannesPfeifer: @rattoma There is again a strange linebreak character here

rattoma:
thank you. should be fixed now

matlab/gsa/scatter_plots.m
rattoma: @JohannesPfeifer thank you. should be fixed now

stepan-a:
Manually merged.

PR #1391, opened 2017-02-09
Make remote parallel computation compatible with new steadystate and …

rattoma:
…auxiliary variable files


stepan-a:
Manually merged.

PR #1395, opened 2017-02-14
Fixes around histval and loglinear

JohannesPfeifer:
- Allow using `histval` to set initial condition for Kalman filter
- Improve compatibility of `loglinear` with `forecast` and `conditional_forecast`-commands (also improves documentation)
- Expands respective integration tests

NB:
1. Design philosophy: the `histval`-block specification should be independent of whether `loglinear` is used or not (as is the case with e.g. `initval` or `steady_state_model`). This requires Dynare taking the log automatically when required. This clashes with `conditional_forecast_path`-block, where the specified controlled values are never transformed.
2. These changes do not affect the `ramsey`-command, which seems to in principle allow for `loglinear` and `histval`. Should we try to make it compatible or provide an error?


stepan-a:
@JohannesPfeifer I like the idea of being able to set the initial condition for the states in the Kalman filter, but:

 - I really do not think that it is a good move to use an already existing interface to do this. We know from experience that it is a source of (not always expected) problems to have a common interface for very different things. Also, what would be the expected behaviour if we have a simulation in the same mod file (before or after the estimation).

 - Such an interface should also allow to set a covariance matrix for the states, and these  initial conditions (for the level and the variance) may depend on expressions (we do not necessarily want to have constant initial conditions during the estimation). Such an interface would be, for instance, very useful if need to estimate the initial condition of the filter.

For these reasons, I think it would be premature to allow the setting of the initial condition of the Kalman filter (recycling `histval`) in 4.5. This needs more thinking and discussion. I will cherry-pick the commits related to the `loglinear` options.

JohannesPfeifer:
@stepan-a I am fine with cherry-picking the bug-fixes for 4.5 and doing the Kalman filter initialization later. 
But regarding your arguments, I am split. 

1. I don't think that this is recycling `histval` for a new purpose. `histval` always has been about setting the value of the historical state variables. I only made the Kalman filter honor this setting. 
2. What is the interaction effect between estimation and simulation in the presence of `histval` that you are worried about? Any `histval`-statement will be kept after estimation. If a user runs `stoch_simul` with `periods` after `estimation`, the starting point of these simulations would still be at the `histval` values, unless one uses `smoother2histval`. That seems a desired behavior.

That being said, I would be OK with having a more encompassing solution that allows for expressions in the initial condition setting and setting the initial covariance.

stepan-a:
Manually merged (partially).

JohannesPfeifer:
For reference: unmerged commits: 298e0ab and 	a8822bc

PR #1396, opened 2017-02-15
smoother2histval: trap cases where both point and posterior estimates are available

rattoma:


rattoma:
I added another commit, this time related to user defined mode_file option which is not preserved in dynare_estimation [re-set to empty].

JohannesPfeifer:
@rattoma I have never used this feature, but when working on #1395 I was wondering how this affects loading smoother results and histval-files? I particularly envision cases where someone switches the `loglinear`-option on or off. In this case, we need to know whether the saved/stored values are in logs or not. #1395 takes the position that `histval` should always store the level, not the logged value and that the log is then taken whenever `loglinear` is specified. Do you know what this implies for the current ticket.

rattoma:
@JohannesPfeifer smoother2histval uses SmoothedVariables and SmoothedShocks. As soon as the loglinear option is properly treated for SmoothedVariables, i.e. SmoothedVarables stores always levels, I think this should imply that the histval generated will always store the levels.
However, looking at store_smoother_results, it seems that SmoothedVariables would instead include logged varibles. Which means histval will provide logged as well... 
Following your reasoning, we should then explicitly allow for a loglinear option in moother2histval, where histval is the exponent of smoothed variables if needed?

JohannesPfeifer:
Exactly. We would need to store an indicator whether loglinear was used or not and then adjust smoother2histval accordingly.

rattoma:
@JohannesPfeifer another small bit added to this PR

stepan-a:
Manually merged.

PR #1399, opened 2017-03-07
Fix check for existence of dname variable

JohannesPfeifer:


JohannesPfeifer:
@houtanb I amended the commit to prevent this issue.

PR #1402, opened 2017-03-10
Add warning if prior has an asymptote

### LINKED ISSUE #1401: Dynare should give warnings when priors are unbounded
It is generally a bad idea to have a prior with unbounded density, as even with reasonably tight identification, such a prior can easily swamp the likelihood. This is particularly problematic when the prior is unbounded at the edges of its support, as then the mode will be driven to the edge, causing further problems with Hessian computation.

At present in Dynare, it is a bit too easy to inadvertently specify such a prior, given the mean and std. dev. specification of the beta distribution. I would suggest that Dynare should give a warning when such are encountered.

For the beta distribution, the prior has bounded support providing alpha>=1 and beta>=1. This means that the variance of the prior must be less or equal to `min(m*(1-m)^2/(1+1-m), (1-m)*m^2/(1+m))`, where `m` is the mean of the distribution.

JohannesPfeifer:
Can happen with beta and gamma distribution. Closes #1401


JohannesPfeifer:
@houtanb I am not sure. Up to this point, we did not issue this as a warning, but rather as information that this is unusual. I would in principle be open to changing this to a warning. That being said, making this into a warning would imply that the `fs2000.mod` we ship in `examples` will trigger a warning whenever someone runs the mod-file. But depending on the decision for #1403, this would not be an issue.

Issue #1407, opened 2017-03-20
Keep track of whether smoother results are in logs and adjust smoother2histval accordingly


JohannesPfeifer:
See #1396

JohannesPfeifer:
@stepan-a Could you please add the bug label?

PR #1408, opened 2017-03-20
Store indicator whether smoother was used with loglinear

JohannesPfeifer:
Required to keep track of whether estimates were logged. Related to #1407


stepan-a:
Manually merged.

PR #1411, opened 2017-03-20
Provide informative error message if non-stationary model is estimate…

### LINKED ISSUE #1377: Decide on treatment of qz_criterium in estimation with particle filter
Currently, if a unit root is present, estimation with `order=2` will result in an error. Using `diffuse_filter` will disable the check, but obviously makes no sense for particle filtering. See http://www.dynare.org/phpBB3/viewtopic.php?f=1&t=13126

JohannesPfeifer:
…d using particle filter

Closes #1377

### DIFF
--- a/matlab/initial_estimation_checks.m
+++ b/matlab/initial_estimation_checks.m
@@ -165,8 +165,15 @@ elseif imag(fval)
 end
 
 if info(1) > 0
+    if DynareOptions.order>1
+        [eigenvalues_] = check(Model,DynareOptions, DynareResults);
+        if any(abs(1-abs(eigenvalues_))<1e-10)
+            error('Your model has a unit root, but estimation at order>1 currently only supports stationary models')
+        end
+    else
     disp('Error in computing likelihood for initial parameter values')
     print_info(info, DynareOptions.noprint, DynareOptions)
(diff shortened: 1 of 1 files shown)


stepan-a:
I am not sure I agree with the message. I have the feeling that we just need to change the initialisation of the filter in this case... I need to check.

JohannesPfeifer:
@stepan-a At the current stage, the message is correct, because we have not implemented a proper initialization for the non-stationary states. I expect to lift this restriction, but not for `4.5`.

stepan-a:
@JohannesPfeifer The initialization with a diagonal matrix is implemented. I need to check that it works.

JohannesPfeifer:
@stepan-a It seems to work, but we do not have an interface in place to select `options_.particle.initialization` (and documentation is missing). 
We also need a way to set `qz_criterium` for the particle filter. Currently. the only way to do this is the `diffuse_filter`-option.

stepan-a:
Fixed. Added option `nonlinear_filter_initialization` in f9a462bf07225f3ca018c6f3bdb97cb0f65d48db. If the model has unit roots, the user must set `nonlinear_filter_initialization=3`.

PR #1421, opened 2017-03-25
utility to decompose initial condition effect

rattoma:
added new utility

### DIFF
--- a/matlab/initial_condition_decomposition.m
+++ b/matlab/initial_condition_decomposition.m
@@ -1,11 +1,11 @@
 function oo_ = initial_condition_decomposition(M_,oo_,options_,varlist,bayestopt_,estim_params_)
-% function z = initial_condition_decomposition(M_,oo_,options_,varlist)
-% Computes shocks contribution to a simulated trajectory. The field set is
-% oo_.shock_decomposition. It is a n_var by nshock+2 by nperiods array. The
-% first nshock columns store the respective shock contributions, column n+1
-% stores the role of the initial conditions, while column n+2 stores the
-% value of the smoothed variables.  Both the variables and shocks are stored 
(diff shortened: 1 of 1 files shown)


houtanb:
@rattoma please `rebase` this on `origin/master`. You will see it doesn't work because the options for `plot_shock_decomposition` are stored in the wrong substructure of `options_`. Also please change the date to 2017 instead of 2009-2016, and fix line 2 to match the function call.

rattoma:
@houtanb thank you. I actually re-based but overlooked the move of options, sorry. should be fixed now.

houtanb:
@rattoma does this need an interface?

rattoma:
@houtanb we may leave this without and undocumented.
Ideally, we could add an interface with only a subset of plot_shock_decomposition options, namely:

- varlist
- type
- detail_plot
- plot_init_date
- plot_end_date
- steadystate
- write_xls

houtanb:
@rattoma what should the defaults be for `plot_init_date` and `plot_end_date`? These have not been set for `plot_shock_decomposition` by the way. Do you want me to add them there too?

rattoma:
@houtanb  should be empty. also for plot_shock_decomp thanks!

houtanb:
@rattoma I pushed the changes in 1891ab70f87561ce5597c9c5202606324d331685. Would the correct description in the manual be:
`plot_init_date` = DATE; If passed, this is the beginning date for the plot.

What is the default value for this option? What is the default value for the end date? By this I mean, not `[]` which is what it is set to, but the date you use in the code?

houtanb:
@rattoma do you want to make the entry in the manual? At the same time you can make the entry for `initial_condition_decomposition`. Please let me know

PR #1424, opened 2017-03-26
Manual changes related shock_decomposition

JohannesPfeifer:
@rattoma Open questions are

1. `oo_.realtime_conditional_shock_decomposition` is set but not documented
2. The description "conditional shock decomposition math{Y(T|T)} conditional on math{Y(T|T-1)},
i{i.e.} math{Y(t|T)} for math{t=[T-1@dots{}T]}." cannot be right. There is something of with the range of t, I guess
3. I am confused by mixture between the information set and forecast horizon. The `realtime_shock_decomposition` manual entry always only conditions on T-1, but it takes a general `forecast` option. It seems there is something missing in the description, most probably with respect to the forecast shock decomposition
4. `WriteShockDecomp2Excel` uses a file called `xlwrite`. Is this a file for MAC one needs to install?
5. The `pooled` is not defined. You refer to `plot_shock_decomposition`, but there is nothing there
6. I don't understand the part about `vintage`. It can take values from 0 to 4. But then you repeat the description of `realtime`, suggesting that the meaning of the four options changes if realtime has a different value

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -1113,6 +1113,9 @@ mutually exclusive arguments are separated by vertical bars: @samp{|};
 @var{INTEGER} indicates an integer number;
 
 @item
+@var{INTEGER_VECTOR} indicates a vector of integer numbers [@var{INTEGER_1} ... @var{INTEGER_N}]
+
+@item
 @var{DOUBLE} indicates a double precision number. The following syntaxes
 are valid: @code{1.1e3}, @code{1.1E3}, @code{1.1d3}, @code{1.1D3}. In
 some places, infinite values @code{Inf} and @code{-Inf} are also allowed;
@@ -7265,7 +7268,7 @@ realtime shock decompositions are computed, @i{i.e.} for
 @anchor{forecast_shock_decomposition} Compute shock decompositions up to
 @math{T+k} periods, @i{i.e.} get shock contributions to k-step ahead forecasts.
 
-@item save_realtime = [@var{integer1} ... @var{integern}]
+@item save_realtime = [@var{INTEGER_VECTOR}]
 @anchor{save_realtime} Choose for which vintages to save the full realtime
 shock decomposition. Default: @math{0}.
 @end table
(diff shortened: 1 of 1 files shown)


rattoma:
@JohannesPfeifer 
1.2.3. indeed this should fixing timing issues in conditional_shock_decomposition. need to update the manual.
4. yes, you are right, xlswrite does nto work on mac. 
5. need to fix
6. Vintage should go to nobs. @houtanb: is 0 to 4 the range allowed by the preprocessor?

JohannesPfeifer:
@rattoma If you tell me what to update, I can attach it to this PR. We could for example discuss necessary changes via Skype.
4, My question was whether `xlwrite` is a function one has to download, i.e. https://de.mathworks.com/matlabcentral/fileexchange/38591-xlwrite--generate-xls-x--files-without-excel-on-mac-linux-win . If yes, we should document this
6. Any integer is allowed on the preprocessor side. Making sure that `vintage` is not bigger than 4 must be done in the Matlab code.

rattoma:
@JohannesPfeifer yes, let's have a skype chat
4. yes, that's exactly the function to download. We may just document it or add it to the missing subfolder?
6. OK. then we can also talk about this via skype

JohannesPfeifer:
@houtanb The PR should be finished now. @rattoma has checked the descriptions.

houtanb:
@JohannesPfeifer @rattoma neither `detail_plot` nor `interactive` take arguments.

houtanb:
@JohannesPfeifer `steadystate` takes an `INT_NUMBER` not a `BOOLEAN`; `1` and `0` are integers even if they're interpreted as booleans. And, actually, this should be changed to not take any argument at all. I'll take care of that.

houtanb:
@JohannesPfeifer `save_realtime = [integer1 ... integern]` should be `save_realtime = [INTEGER_VECTOR]` (`INTEGER_VECTOR` needs to be defined in section `4.1`

JohannesPfeifer:
@houtanb Thanks. I updated the commit and defined `integer_vector`. Note that we have regularly used `BOOLEAN` in the manual to describe a command taking only 0 or 1.

PR #1427, opened 2017-03-30
- test the rank of Pinf in univariate smoother for the subset of obse…

rattoma:
…rvables, in order to trap cases where this is not already trapped in compute_Pinf_Pstar.m

- issue warning when diffuse steps never end [maybe we should issue an error?]


houtanb:
@rattoma this change breaks `analytic_derivatives/fs2000_analytic_derivation.mod`. Please fix the commit or the test.

Issue #758, opened 2014-10-13
Write specialized code for loading data...


stepan-a:
We should not use the same code if we load data with the "legacy approach" (`datafile` option of the `estimation` command) or if we use the `data` command. In the first case we should create a `dseries` restricted to the variables listed in `options_.varobs`. See PR #741.

stepan-a:
Fixed in 63d31572677315705978ac507b87ef2f7ce8c74f.

PR #1443, opened 2017-04-27
Add legacy function for loading mat-files with old interface

### LINKED ISSUE #1364: Allow loading data from mat-files that contain other variables
We rely on `load_mat_file_data.m` from `dseries` to read in `mat`-files. There we use
```
for i=1:length(varlist)
    try
        tmp = getfield(datafile,varlist{i});
        data = [data,  tmp(:)];
    catch
        error(['load_mat_file:: All the vectors (variables) in ' inputname(1) ' must have the same number of rows (observations)!'])
    end
end
```
to read _all_ variables and concatenate them in an array. If there is any other variable in the mat-file that does not have conformable dimensions, we abort with a crash. This deviates from earlier behavior and will result in many users reporting problems with backward compatibility if we release `4.5` this way. Is there a way to only load the specified `varobs`? Currently, it seems impossible to pass this through the `dseries`.

JohannesPfeifer:
Follows approach of #758 and closes #1364

### DIFF
--- a/matlab/load_mat_file_data_legacy.m
+++ b/matlab/load_mat_file_data_legacy.m
@@ -1,0 +1,41 @@
+function data  = load_mat_file_data_legacy(datafile, varobs)
+
+% Copyright (C) 2017 Dynare Team
+%
+% This file is part of Dynare.
+%
+% Dynare is free software: you can redistribute it and/or modify
+% it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


stepan-a:
Manually merged (with bug fix in e1c983cc7b9be18c6f79b1258d87b59818e4ea7b).

PR #1464, opened 2017-06-05
Add missing check on whether bayestopt_ is empty

JohannesPfeifer:


stepan-a:
@JohannesPfeifer Sorry I have just pushed a fix. There was also a problem with xparam1 being undefined.

PR #1470, opened 2017-06-16
Fix bug in use_shock_groups

JohannesPfeifer:
Option was not written to correct subfield of options_-structure. Related to 7cd56e96877de4ed2dd22bb579884ab292034d71

### DIFF
--- a/matlab/plot_shock_decomposition.m
+++ b/matlab/plot_shock_decomposition.m
@@ -30,7 +30,6 @@ function [z, steady_state] = plot_shock_decomposition(M_,oo_,options_,varlist)
 
 options_.nodisplay = options_.plot_shock_decomp.nodisplay;
 options_.graph_format = options_.plot_shock_decomp.graph_format;
-options_.use_shock_groups = options_.plot_shock_decomp.use_shock_groups;
 
 % indices of endogenous variables
 if size(varlist,1) == 0
--- a/preprocessor/DynareBison.yy
+++ b/preprocessor/DynareBison.yy
@@ -2553,8 +2553,8 @@ plot_shock_decomposition_options_list : plot_shock_decomposition_option COMMA pl
                                       | plot_shock_decomposition_option
                                       ;
 
-plot_shock_decomposition_option : o_psd_use_shock_groups
-                                | o_psd_colormap
+plot_shock_decomposition_option : o_use_shock_groups
+                                | o_colormap
                                 | o_psd_nodisplay
                                 | o_psd_graph_format
                                 | o_psd_detail_plot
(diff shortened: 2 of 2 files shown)


houtanb:
@JohannesPfeifer are you sure about this change? `plot_shock_decomposition` uses `o_psd_use_shock_groups` which prints the option to `plot_shock_decomp.use_shock_groups`.

`o_use_shock_groups` is used by `shock_decomposition` and `realtime_shock_decomposition` not by `plot_shock_decomposition`

JohannesPfeifer:
@houtanb I would like @rattoma to confirm the correctness, but I am pretty sure. `shock_decomposition` internally relies on `plot_shock_decomposition`. Having the option in a different field means that the latter cannot use it.

rattoma:
@JohannesPfeifer @houtanb 
Indeed plot_shock_decomposition is always used within `shock_decomposition` and `realtime_shock_decomposition` as factorized plotting utility. 

I do not recall why we needed the extra field `options_.plot_shock_decomp` on top of `options_.shock_decomp`, but the patch proposed by @JohannesPfeifer is correct, since plot_decomposition would look at its own list of options ... thank you

houtanb:
@rattoma @JohannesPfeifer if this is the case, then there are two things:
1) We should delete the option `o_psd_use_shock_groups` and replace it by `o_use_shock_groups`
1) I don't understand then the change in `matlab/global_initialization.m`. If this field is not used why not remove it completely?

rattoma:
I may start recalling: the reason should be to avoid the plot options to be persistent, which suggests that the pre-processor should always re-set to default all plot options before any new call to `plot_shock_decomposition`.
If this is the case, then we have to keep the current layout.

houtanb:
This is what this sort of change is for. So that the option passed to the various commands are not persistent.....

rattoma:
yes, which means this PR is fine. many thanks

houtanb:
Ah, so I had misunderstood. Now you are fine with the persistence of options and will not change the backend? If this is the case, then @JohannesPfeifer should make the changes outlined here: https://github.com/DynareTeam/dynare/pull/1470#issuecomment-309430672

JohannesPfeifer:
@houtanb Now I am confused what needs to be done

houtanb:
@JohannesPfeifer the changes that are in this comment. @rattoma is ok with the idea of merging these two options. But, the change you implement in f595d37 has two options, `o_psd_use_shock_groups` and `o_use_shock_groups` that do the same thing.

Also, I don't understand the change you made to `global_initialization.m`. Why do we need `options_.shock_decomp.use_shock_groups` if we always reference `options_.plot_shock_decmp.use_shock_groups`?

rattoma:
@houtanb no sorry, I'm OK with NOT having persistent options for plot_shock_decomposition

houtanb:
@JohannesPfeifer are the changes to be made here clear or would you like me to make the changes?

JohannesPfeifer:
@houtanb I gave it another try

houtanb:
@JohannesPfeifer your change looks good. Can we also remove line 33 of `plot_shock_decomposition.m`? It seems useless.....

JohannesPfeifer:
@houtanb Done

PR #1475, opened 2017-07-05
Posterior moments: fix bug that prevented updating decision rules for…

JohannesPfeifer:
… parameter vector, leading to wrong results/crashes when computing second moments

When removing globals in 24cd4236710b452f5131b20d844a5bba991c0d3b the call to set_parameters.m, which relies on M_ being global, was not removed. The problem arises

1. When computing second moments for big models with drsize*SampleSize>MaxMegaBytes (in which case decision rules dr were not saved, but recomputed)
2. When computing the conditional variance decomposition for all models regardless of size (dsge_simulated_theoretical_conditional_variance_decomposition.m relied on the wrong M_.Sigma_e in this case)


stepan-a:
Manually merged.

PR #1485, opened 2017-08-01
Provide meaningful error message if oo_.exo_simul is larger than opti…

JohannesPfeifer:
…ons_.periods

Related to https://forum.dynare.org/t/dynare-4-5-1-issue-with-deterministic-simulation/10450


JohannesPfeifer:
@stepan-a This is definitely a bug. In `4.4.3` we simply ignored anything happening after `options_.periods`. Now Dynare crashes. Either we provide a correct error message as is proposed here or we go back to the behavior of Dynare 4.4.3, but let the preprocessor check whether this case happens and provide an error unless `nostrict` is passed.

stepan-a:
Manually merged.

PR #1488, opened 2017-08-11
DsgeSmoother.m: provide informative error instead of crashing when mo…

JohannesPfeifer:
…del cannot be solved


stepan-a:
Manually merged.

PR #1491, opened 2017-08-16
Fix plotting of measurement errors in non-MCMC case

JohannesPfeifer:
Number of plots was wrongly computed


stepan-a:
Manually merged.

PR #1498, opened 2017-08-29
Reset order=1 for all of stoch_simul.m if the Hessian is 0

JohannesPfeifer:
Prevents various crashes where functions called rely on fields existing at order>1


stepan-a:
Manually merged.

PR #1499, opened 2017-08-29
Make empirical moments independent of simul_replic by outputting firs…

### LINKED ISSUE #1487: Potentially make empirical moments independent of simul_replic
We currently rely on the last simulated series for computing moments. If `simul_replic` is increased, the moments will therefore change. I would propose to output the first instead of the last run in `simult.m`. That way, the results would always be the same. Of course, this would entail some loss of backward compatibility. I would accept this, because in principle any replication for the moments should do and the manual states that `simul_replic` is not used for moment computation. Related to https://forum.dynare.org/t/impact-of-simul-replic-option-on-empirical-moments/10553/4

JohannesPfeifer:
…t instead of later replication

Closes #1487


stepan-a:
Manually merged.

Issue #1400, opened 2017-03-08
Investigate dseries problems in parallel execution


JohannesPfeifer:
When executing the attached file, which runs the ```UseParallel``` feature of ```fmincon```, a crash results because the dseries data is transformed to a structure:
```
Warning: Element(s) of class 'dseries' do not match the current constructor definition.  The element(s) have been converted to structures.
> In parallel.internal.pool.deserialize (line 9)
  In parallel.internal.pool.deserializeFunction (line 12)
  In remoteParallelFunction (line 33)
Warning: Element(s) of class 'dates' do not match the current constructor definition.  The element(s) have been converted to structures.
> In parallel.internal.pool.deserialize (line 9)
  In parallel.internal.pool.deserializeFunction (line 12)
  In remoteParallelFunction (line 33)
```

[SW.zip](https://github.com/DynareTeam/dynare/files/827628/SW.zip)

tholden:
I have encountered this issue previously too. I recall hacking in a fix to my dynare and dseries forks (quite a while ago), but I doubt it's very elegant.

stepan-a:
We have two branches in the `dseries` repository. In Dynare/master we use the branch where the class is implemented with the old OOP interface. The master branch (of `dseries`) uses instead the new OOP interface. I need to test your file with this to see if the master branch fixes the issue. This branch is still not used in Dynare because of a bug in the implementation of the new interface in Octave.

Issue #1051, opened 2015-09-03
Eliminate the global variable objective_function_penalty_base


tholden:
This causes problems when the estimation function is evaluated in parallel, using fmincon's 'UseParallel','always' or similar. Further problems with parallel objectives are given in this issue: https://github.com/DynareTeam/dseries/issues/10

stepan-a:
Yes this is the last global in ```dsge_likelihood```. We have at least to precise in the manual that this code cannot be called in parallel. I think it would be a lot of work to pass this variable as an input argument. Maybe we could put the global in a if statement, but I don't know what would be the cost of this solution. Another possibility is to write a version of ```dsge_likelihood```without endogenous penalty (which is only used with ```mode_compute=4,5```.

JohannesPfeifer:
Is there any reason we do not put it as a field into one of the structures we already use? For example into options?

tholden:
dsge_likelihood ought to support being called in parallel really. I've been using CMAES on a multi-modal objective function, with the main loop in parallel. Without the parallelisation it would really be prohibitively slow. As a work-around, I have:
```
if isempty( objective_function_penalty_base ) || objective_function_penalty_base <= 1
	objective_function_penalty_base = Inf;
end
if isstruct( DynareDataset )
    dseries( 'initialize' );
    DynareDataset = dseries( DynareDataset );
end
```
There's also a required change to dseries to get this to work, to support initializing from a struct.

JohannesPfeifer:
I fully agree, but this will be tricky. I am willing to do the change, but we need to find out how to do this. The penalty is sometimes updated. By being global this update is available in later calls. When having the penalty as an input, one needs to make sure that any update is passed to later calls. But this path dependeny is fundamentally incompatible with a parallel context. Maybe we should think about phasing this change in penalties out. @stepan-a @MichelJuillard what exactly is the purpose of this penalty?

stepan-a:
We cannot remove the penalty for ```mode_compute=4,5```. I think that it should be possible to pass it as an argument but it's not simple because we do not systematically update its value (for instance we do not update the penalty when calling numgrad routines (so we would have to pass also a flag to indicate if the penalty bas has to be updated).

The endogenous penalty helps the gradient based methods to find a smooth escape from problematic regions (where BK conditions are not satisfied, a real steady state does not exist, ...).

JohannesPfeifer:
[...]

PR #1508, opened 2017-09-07
Optimization parallel support

tholden:
Fixes https://github.com/DynareTeam/dynare/issues/1400
Also provides parallel support for CMAES.

See also my related pull requests to dates and dseries:
 * https://github.com/DynareTeam/dseries/pull/26
 * https://github.com/DynareTeam/dates/pull/9

### DIFF
--- a/matlab/optimization/parallel_wrapper.m
+++ b/matlab/optimization/parallel_wrapper.m
@@ -1,18 +1,9 @@
 function RV = parallel_wrapper( objective_function, XV, varargin )
     n = size( XV, 2 );
     RV = zeros( 1, n );
-    
-    global objective_function_penalty_base
-
-    if isempty( objective_function_penalty_base ) || objective_function_penalty_base <= 1
-        objective_function_penalty_base = Inf;
-    end
-    
-    in_objective_function_penalty_base = objective_function_penalty_base;
-        
+          
     parfor i = 1 : n
         R = [];
-        SetGlobals( in_objective_function_penalty_base );
         WarningState = warning( 'off', 'all' );
         try
             R = objective_function( XV( :, i ), varargin{:} ); %#ok<PFBNS>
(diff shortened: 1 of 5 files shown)


matlab/optimization/parallel_wrapper.m
```diff
@@ -0,0 +1,36 @@
+function RV = parallel_wrapper( objective_function, XV, varargin )
+    n = size( XV, 2 );
+    RV = zeros( 1, n );
+    
+    global objective_function_penalty_base
```
JohannesPfeifer: @tholden We removed this global variable in  #1051. Why are you reintroducing it?

tholden: Because I hadn't noticed/had forgotten that it had been removed! Fixed in the commit below.

JohannesPfeifer:
@tholden We are very sorry this got lost. We were too busy getting 4.5.2 out. The main issue was that that there had been a massive change to `dates` and `dseries` outstanding for the unstable branch that has now been merged. This is what creates the merge conflict. Do you by any chance know whether these changes resolved the required changes for the submodules?

Apart from this, I wanted to have a look at this interesting PR for quite some time, but teaching commitments intervened. The issue I am having is that you did not document anything here, Thus, I have a hard time following what you are actually doing and how I could test this. In particular, I do not know how to invoke a parallel mode-finding.

tholden:
You've switched from the old-oop-style branch? The master branch requires different fixes I seem to remember. I'm afraid I may not be able to update this till December as I've got two papers to finish by the end of the month for a funder.
It is set up so parallel CMAES is invoked using 'UseParallel', just as it is for any other solver. See: https://github.com/DynareTeam/dynare/pull/1508/commits/bdfb68fa7fb488f254f2975beac1920b6eea455e
But yes, I should have explained this in the original PR.

ivanlozev:
I would like ask what has happened to this branch, a sit is closed now? Will it be merged in the next Dynare version or it is discontinued? I am asking, because I want to implement simple GMM through fmincon, but I am having trouble running fmincon through MultiStart with the option ‘UseParallel’, true. This problem seems to be connected to the elements of class ‘dates’ being converted to structures.  Is there a workaround for that or I better wait for a fix in Dynare?

tholden:
I have a stable Dynare release with this feature available from here: https://github.com/tholden/dynare/releases

PR #1517, opened 2017-09-15
mode_check.m: fix case where parameter is exactly 0

JohannesPfeifer:
In this case, interval length was 0, crashing the plotting


stepan-a:
Manually merged.

Issue #1409, opened 2017-03-20
Add an interface for setting initial condition of the Kalman filter


stepan-a:
See discussion in #1395.

JohannesPfeifer:
We need a `filter_initial_state` block that follows the syntax of the `observation_trends` which was
``` 
observation_trends;
P_obs (g_p);
Y_obs (g_y);
end;
``` 
The difference is that now we only allow setting state variables as in `histval`-blocks (same consistency checks apply). Within the bracket, we should allow for any valid expression of constants and parameters as for `observation_trends`. 

The biggest question is the structure for storing these expressions. I would go for a `N by 2` cell array in `M_.filter_initial_state`. I prefer this to putting it in  `options_` (although we already store the trend coefficients in `options_.trend_coeff`). The `N by 2` cell should contain the name of the state in the first column and the expression in the second. Along rows it should be in the same order as `M_.endo_histval`. This should also allow us to keep track of auxiliary variables introduced for lags>1. 

Checks on whether the expression provided works belong into `initial_estimation_checks`

houtanb:
@stepan-a can this be done for 4.5 or should we wait for 4.6?

JohannesPfeifer:
@houtanb @stepan-a I don't think it's realistic to have this fully implemented for 4.5. Nevertheless, I would like to continue to work on this once the preprocessor part is done.

houtanb:
@JohannesPfeifer ok, I'll do the preprocessor part this week

houtanb:
@JohannesPfeifer in `histval` blocks we do not have consistency checks regarding the type of variable set. The only checks we do are made when `all_values_passed` has been set. In this case, we ensure that all endogenous and exogenous variables have been assigned in `histval`...

JohannesPfeifer:
@houtanb When you run
```
// Tests the histval block in deterministic setup
// In particular test if it works on endos and exos substituted by an aux var

var c k z;
varexo x;

parameters alph gam delt bet aa;
alph=0.5;
gam=0.5;
delt=0.02;
bet=0.05;
aa=0.5;


model;
c + k - aa*x*k(-1)^alph - (1-delt)*k(-1);
z = 0.9*z(-2);
c^(-gam) - (1+bet)^(-1)*(aa*alph*x(-1)*k^(alph-1) + 1 - delt)*c(+1)^(-gam);
end;

initval;
x = 1;
k = ((delt+bet)/(1.0*aa*alph))^(1/(alph-1));
[...]

PR #1522, opened 2017-09-25
Allow setting initial state of Kalman filter

JohannesPfeifer:
Related to #1409 and #1395


stepan-a:
@JohannesPfeifer and @houtanb Some of the problems/questions I have with this PR:

1. The function `set_Kalman_smoother_starting_values` seems to be defined twice (in `DsgeSmoother` and `dsge_likelihood`). The function should be factorized in a separate file.

2. I do not like the idea of having the initial condition in a `M_`, it would be much simpler to generate an m file returning a vector. Calls to `eval` are in general prohibitive.

3. I may have missed something, but does this PR take care of the derivatives w.r.t. the parameters?

4. Why should Dynare fail if the user specify an initial condition for non predetermined variables?  It seems to be that this behaviour is inconsistent with what we do in other situations (e.g. perfect foresight simulations).

5. It would be good to have a test with more than one lag... To check that there is no issues with auxiliary variables.

JohannesPfeifer:
@stepan-a 
1. I initially planned on doing so, but the variable ordering is different in the two files (one is for the filter, one for the smoother), making the two `set_Kalman*_starting_values`-functions different and local to their caller.
2. I followed the approach we already do with `observation_trend`, but would be all for the more non-`eval`- version. My proposal is to first implement it this way and then get rid of the `eval`-calls here and in the trend setting in a separate ticket
3. You mean for checking identification? No, we don't do that and I think we have the same issue with `observation_trend`
4. Because there is no situation where you would like to set non-predetermined variables in this block. For perfect foresight simulations,  `initval` doubles as the initial values for the solver (which is not ideal as discussed in http://www.dynare.org/DynareWiki/DeterministicSimulationBlocks). But there is no such functionality for the `filter_initial_state`-block. Thus, I think we should protect users from mistakes. We already do the same with histval, where we only allow setting predetermined variables for the lags where they appear.
5. I agree and will work on that.

PR #1526, opened 2017-10-03
Fix get_posterior_parameters

JohannesPfeifer:
Get rid of setting variables from the base workspace in the function, making the function read-only (as the name suggests)


stepan-a:
Manually merged.

PR #1528, opened 2017-10-03
Add occbin folder to path

JohannesPfeifer:


houtanb:
@JohannesPfeifer can you make the `occbin` changes in your repository and then make a pull request when everything is ready? We can delete the `occbin` branch in `dynare.git` because it should never have been there to begin with.

PR #1529, opened 2017-10-04
dynare_estimation_init.m: Remove call to check_list_of_variables

JohannesPfeifer:
Already done in dynare_estimation.m


stepan-a:
Won't merge because of an issue with GSA. I pushed a fix in 6e656fbf50402bb70dbd87e2fdc07d30a919ccce instead.

PR #1553, opened 2017-11-14
- bug fix for when NO parameter is detected below pvalue threshold

rattoma:
- fix to the case when no plot is done when there are too few behavioral/non-behavioral runs [as it already happens when one of the two sets is empty]
- added more explicit messages on screen


stepan-a:
Manually merged.

PR #1559, opened 2017-11-22
AnalyseComputationalEnvironment.m: print out command that caused erro…

JohannesPfeifer:
…r as well as system response


stepan-a:
Manually merged.

PR #1560, opened 2017-11-22
Filter out cases where stochastic_solvers.m is called at higher order…

JohannesPfeifer:
… without k_order_solver


stepan-a:
Manually merged and cherry-picked in 4.5 branch.

PR #1552, opened 2017-11-14
Shock decomp

rattoma:
This pull request contains a number of bug fixes to the new plottting and realtime  shock decompositions.
All bug fix commits should also be cherry picked to 4.5
there is also a new option `fast_realtime` that should apply to
`options_.shock_decomp.fast_realtime`
for the 
`realtime_shock_decomposition`
command.

PR #1563, opened 2017-11-22
Fast realtime

JohannesPfeifer:
Add `fast_realtime` option for #1552

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -7539,6 +7539,11 @@ realtime shock decompositions are computed, @i{i.e.} for
 @item save_realtime = @var{INTEGER_VECTOR}
 @anchor{save_realtime} Choose for which vintages to save the full realtime
 shock decomposition. Default: @math{0}.
+
+@item fast_realtime
+@anchor{fast_realtime} Runs the smoother only twice: once for the last in-sample 
+and once for the last out-of-sample data point. Default: not enabled.
+
 @end table
 
 @outputhead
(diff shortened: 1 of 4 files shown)


houtanb:
Can be merged once #1552 is merged

stepan-a:
@JohannesPfeifer I rebased this PR on master, and the test you added failed (the rebased branch is available here [pr#1563](https://mnemosyne.adjemian.eu/stepan-a/dynare/tree/pr%231563)). I don't see any problem in your PR. Did you successfully run `ls2003_plot`? If yes it means that I probably made something wrong when I merged #1552...

stepan-a:
... Or that there is something wrong in the implementation (by @rattoma) of the option in the first place.

JohannesPfeifer:
@stepan-a No, this seems to be a Matlab version issue that hid a bug. In the last run, there is only one observation, so the data mean computed is not along the first, but the second dimension, because the data is just a vector instead of a matrix. This results in the data mean being a scalar instead of a 5 by 1 vector. You need to also add https://github.com/JohannesPfeifer/dynare/commit/2d371b1997f5fa07fcbbf47e5923d7817d07c6b9 to the PR (or merge that whole branch at https://github.com/JohannesPfeifer/dynare/tree/1563)

JohannesPfeifer:
@rattoma Thinking about it, the line 
```options_.nobs=fast_realtime;``` 
in `realtime_shock_decomposition` seems problematic. If I understand it correctly, you want to run the smoother on the last in-sample observation. But 
1. That way, there will be 1 observation only after the first observation, so that is not the last one, but the first one.
2. Using the `prefilter` option implies that the data mean that is subsequently added is also only base on this one observation. Is this desired?

stepan-a:
Manually merged. @JohannesPfeifer Can you copy your last questions/remarks to @rattoma in a new issue?

rattoma:
@JohannesPfeifer indeed. The use I have foreseen for the fast_realtime option is that it should be either 0 or the last in sample observation you want to use for this option.
Then the user input should be a meaningful nobs, not really 1!

PR #1574, opened 2017-12-01
Make Octave error message more constructive and informative

JohannesPfeifer:
See the discussion at http://savannah.gnu.org/bugs/?52569


houtanb:
@JohannesPfeifer Can you add the link to the conversation you had with the Octave developers in the comment? That could be useful down the line; not sure we'd think to come look at the PR....

JohannesPfeifer:
@houtanb Done.

PR #1582, opened 2018-01-17
further bug fixes in Shock decomp

rattoma:
both fixes should be cherry picked to 4.5.
48fa05b needs manual cherry pick since names list became cell arrays in 4.6

### DIFF
--- a/matlab/expand_group.m
+++ b/matlab/expand_group.m
@@ -34,16 +34,19 @@ mydata=get(findobj(gcf,'tag',['group' int2str(ic)]),'userdata');
 if isfield(mydata,'shock_decomp')
     options.shock_decomp=mydata.shock_decomp;
 end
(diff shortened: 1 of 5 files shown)


JohannesPfeifer:
@rattoma There is an undocumented option `expand` in the files. What does it do? Should we add a preprocessor option and document it?

rattoma:
@JohannesPfeifer expand is the feature activated by the interactive option of plot_shock_decomposition.
WIth interactive mode, when you use gfroups of shock, there is a context menu available by right-click on each line of the legend of the plot that allow to expand the group, if one needs to check the effect of each indivual shock of the group.
With detailed plot mode, the ui menu it triggered by right-clik on each subplot.
So, this is not really a new option, but indeed this ui menu may be documented in the manual

stepan-a:
Manually merged and cherry picked in 4.5 branch.

PR #1589, opened 2018-01-31
Discretion: Added catch for infinities

JohannesPfeifer:
Comment: A user provided a file in which the Sylvester equation solved using doubling can evaluate to inf. When Doubling fails, one can try the Hessenberg-Schur algorithm.


stepan-a:
Manually merged and cherry picked in 4.5.

PR #1591, opened 2018-01-31
prior bounds: make sure that lower and upper bounds only affect mode-…

JohannesPfeifer:
…finding

https://github.com/DynareTeam/dynare/commit/99dbc8c74d0ecf4c7a610c4a394fd85028cf2741 introduced a bug by disabling the resetting of the prior bounds used for the MCMC. The manual clearly states the bounds are only operational during mode-finding, but not during MCMC (and therefore prior sampling)

### DIFF
--- a/matlab/dynare_estimation_1.m
+++ b/matlab/dynare_estimation_1.m
@@ -422,6 +422,7 @@ end
 
 if (any(bayestopt_.pshape  >0 ) && options_.mh_replic) || ...
         (any(bayestopt_.pshape >0 ) && options_.load_mh_file)  %% not ML estimation
+    bounds = prior_bounds(bayestopt_, options_.prior_trunc); %reset bounds as lb and ub must only be operational during mode-finding
     outside_bound_pars=find(xparam1 < bounds.lb | xparam1 > bounds.ub);
     if ~isempty(outside_bound_pars)
         for ii=1:length(outside_bound_pars)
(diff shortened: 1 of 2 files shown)


JohannesPfeifer:
1. In master, this PR requires https://github.com/DynareTeam/dynare/pull/1590
2. Needs to be cherry picked to 4.5

stepan-a:
Manually merged and cherry picked in 4.5 branch.

rattoma:
Concerning posterior sampling (and on cascade prior sampling), why not adding an option to estimation:

`mh_use_mode_finding_bounds=[INTEGER]`

with default 0. when we set it =1, the line uncommented in this last PR should be trapped by an if clause:
```
if ~options_.mh_use_mode_finding_bounds`
bounds = prior_bounds(bayestopt_, options_.prior_trunc);
end
```

what do you think?

stepan-a:
@rattoma Why would you be interested in having these bounds in the MCMC? Or is it just for the prior sampling?

JohannesPfeifer:
@rattoma @stepan-a I see the need for the implementation of proper truncated priors in Dynare. But for the time being, I would like to avoid introducing hacks that allow to only partially replicate the behavior achieved with proper truncated priors.

rattoma:
@stepan-a @JohannesPfeifer indeed I think about truncated priors.
@JohannesPfeifer  why would this `mh_use_mode_finding_bounds` such a bad thing? it is an explicit option that could be then selected by the user [not hidden as it was currently]. We would add a warning the manual and also on the command line that this would imply wrong estimation of the marginal likelihood, due to the normalization.

JohannesPfeifer:
@rattoma I would like to avoid having a hack like that and instead provide a way to specify a truncated prior where the truncation is correctly taken into account, i.e. where we deal with the normalization.

PR #1592, opened 2018-02-01
small bug fixes

rattoma:
- dynare.texi
- makedataset.m
- smoother2histval.m

### DIFF
--- a/doc/dynare.texi
+++ b/doc/dynare.texi
@@ -7532,9 +7532,9 @@ model).
 (@xref{plot_shock_decomposition}).
 
 @item presample = @var{INTEGER}
-@anchor{presample_shock_decomposition} First data point from which recursive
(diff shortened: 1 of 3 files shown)


JohannesPfeifer:
@rattoma Do all three commits need to be cherry-picked for 4.5?

stepan-a:
I didn't push yet, but I only cherry picked the two first commits. I did not understand that the last one is a bug fix.

rattoma:
@JohannesPfeifer @stepan-a on the last commit: if I am the only one who replaces smoothed variables with updated ones to get faster the correct starting point for pseudo-realtime forecast exercises, then I am the usual hacker ... 
otherwise, this is in principle a bug because if one provides in input or in a file the initializiation of all variables including auxiliary ones without bothering of the consistency of the (-2) ... (-k) data points, then `smoother2histval`  overwrites the user defined info with something different.

PR #1594, opened 2018-02-06
sim1_linear.m: change hard-coded tolerance to option

JohannesPfeifer:
The tolerance for the steady state check should depend on the accuracy of the steady state computation


stepan-a:
Manually merged.

PR #1623, opened 2018-06-29
TaRB: fix bug where incorrect last posterior was returned if last dra…

JohannesPfeifer:
…w was rejected


stepan-a:
Manually merged.

